PR #3, opened 2023-10-09
improve generated event classes

Muinez:
add setters
add constructor


roflmuffin:
Looks great, thanks for the contribution. Would you be able to reset the [API.cs](https://github.com/roflmuffin/CounterStrikeSharp/pull/3/files#diff-e1a65df00ba4a39e8e47c917e86f20d0729c816aea14fe8b1df89af07b74a5f9) file in this PR as it should not have changed based on the diff?

PR #8, opened 2023-10-21
Add helper function FindAllEntitiesByDesignerName

zonical:
The goal of this PR is to add a function that allows plugins to grab a list of all entities with a certain design name so they can be easily manipulated. The example I've used is setting all of the player's money value to 1337.

This could be optimized, but this will require further work internally which I'm not confident with. Right now the function loops through all 16k entity indexes, but this can probably be cut down using the method [CS2Fixes uses by navigating through entity pointers until they run into a null pointer](https://github.com/Source2ZE/CS2Fixes/blob/main/src/utils/entity.cpp#L29). It would be possible to do the basic functionality on the C++ side, but I prefer the design of returning a List, and returning an array from the C++ side isn't easy as I've been told. A feature I would like to in-cooperate into this function is passing through a type (e.g `Utilities.FindAllEntitiesByDesignerName<CCSPlayerController>("cs_player_controller")`) so you can just manipulate all of the entities in the list straight away. I tried doing this, but ran into errors related to `new T()`. My C# knowledge isn't strong so I've abandoned this idea for now and instead return a `List<CEntityInstance>`.

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/API.cs
+++ b/managed/CounterStrikeSharp.API/Core/API.cs
@@ -426,6 +426,16 @@ namespace CounterStrikeSharp.API.Core
 			}
 		}
 
+        public static IntPtr GetFirstActiveEntity(){
+			lock (ScriptContext.GlobalScriptContext.Lock) {
+			ScriptContext.GlobalScriptContext.Reset();
+			ScriptContext.GlobalScriptContext.SetIdentifier(0x3E50DC41);
+			ScriptContext.GlobalScriptContext.Invoke();
+			ScriptContext.GlobalScriptContext.CheckErrors();
+			return (IntPtr)ScriptContext.GlobalScriptContext.GetResult(typeof(IntPtr));
+			}
+		}
+
         public static void HookEvent(string name, InputArgument callback, bool ispost){
 			lock (ScriptContext.GlobalScriptContext.Lock) {
 			ScriptContext.GlobalScriptContext.Reset();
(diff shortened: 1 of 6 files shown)


roflmuffin:
I think it might be better to make this return an IEnumerable rather than a list, so that a consumer can abort the iteration early if they find the entity they are looking for

zonical:
The offset stuff wasn't as hard as I thought. Currently not getting all the entities I want but that's probably because I'm starting at worldent (0) and not the first active entity like CS2Fixes does. I'll see if I can grab the first active entity with a NativeAPI call.

image

zonical:
image
Progress?

roflmuffin:
I think there is also the class version of prev and next which might be better for this use case?

```
CEntityIdentity* m_pPrevByClass; // 0x68	CEntityIdentity* m_pNextByClass; // 0x70
```

roflmuffin:
Also with regard to the IEnumerable, there is a yield return syntax here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/yield

This lets us stop iterating when the enumeration stops

zonical:
> I think there is also the class version of prev and next which might be better for this use case?
> 
> ```
> CEntityIdentity* m_pPrevByClass; // 0x68	
> 
> CEntityIdentity* m_pNextByClass; // 0x70
> ```

I think it would be better to just use the normal Prev and Next as I've designed the function to be able to grab different entities with a `Contains` check on the designer name (e.g if I wanted all of the `cs_` specific entities, I can pass that through into the function and then sort in the calling function). No harm in throwing the pointers into the class though.

> Also with regard to the IEnumerable, there is a yield return syntax here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/yield
> 
> This lets us stop iterating when the enumeration stops

This is funky, but interesting! Will switch it over to use this.

PR #14, opened 2023-10-29
Add css, css_plugins commands.

zonical:
This PR adds two new commands: `css`, and `css_plugins`.
- `css` displays credits (mainly towards roflmuffin) and points towards the acknowledgements file.
- `css_plugins` handles plugin loading, unloading, reloading, and displaying a list of plugins.
I've also moved a few things around in `GlobalContext`: renamed `LoadAll` to `InitGlobalContext`, and split some functions into smaller ones.

I decided to deviate from the MetaMod/SourceMod style of "self-command" and split plugins into it's own isolated category of commands. I did this for future development, like possibly incooperating `apfelwurm` 's [update/install commands](https://discord.com/channels/1160907911501991946/1160907912445710481/1166516777925222400) into it's own category.

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/GlobalContext.cs
+++ b/managed/CounterStrikeSharp.API/Core/GlobalContext.cs
@@ -122,7 +122,6 @@ namespace CounterStrikeSharp.API.Core
 
             foreach (var path in filePaths)
             {
-                if (path.Contains("disabled")) continue;
                 try
                 {
                     LoadPlugin(path);
--- a/managed/CounterStrikeSharp.API/Core/GlobalContext.cs
+++ b/managed/CounterStrikeSharp.API/Core/GlobalContext.cs
@@ -54,6 +54,203 @@ namespace CounterStrikeSharp.API.Core
                 plugin.Unload();
             }
         }
+        public void InitGlobalContext()
+        {
+            Console.WriteLine("Loading GameData");
+            GameData.Load(Path.Combine(rootDir.FullName, "gamedata", "gamedata.json"));
+
+            for (int i = 1; i <= 9; i++)
+            {
+                AddCommand("css_" + i, "Command Key Handler", (player, info) =>
+                {
+                    if (player == null) return;
+                    var key = Convert.ToInt32(info.GetArg(0).Split("_")[1]);
+                    ChatMenus.OnKeyPress(player, key);
+                });
+            }
+
+            Console.WriteLine("Loading C# plugins...");
+            int pluginCount = LoadAllPlugins();
+            Console.WriteLine($"All managed modules were loaded. {pluginCount} plugins loaded.");
+
+            RegisterPluginCommands();
+        }
+
+        public void LoadPlugin(string path)
+        {
+            var plugin = new PluginContext(path);
+
+            try
+            {
+                foreach (var existingPlugin in _loadedPlugins)
+                {
+                    if (plugin.Name == existingPlugin.Name)
+                    {
+                        throw new Exception("Plugin is already loaded.");
+                    }
+                }
+                _loadedPlugins.Add(plugin);
+                plugin.Load();
+            }
+            catch (Exception e)
+            {
+                _loadedPlugins.Remove(plugin);
+                throw;
+            }
+        }
+
+        public int LoadAllPlugins()
+        {
+            DirectoryInfo modules_directory_info;
+            try
+            {
+                modules_directory_info = new DirectoryInfo(Path.Combine(rootDir.FullName, "plugins"));
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(
+                    "Unable to access .NET modules directory: " + e.GetType().ToString() + " " + e.Message);
+                return 0;
+            }
+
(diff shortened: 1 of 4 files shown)


managed/CounterStrikeSharp.API/Core/GlobalContext.cs
```diff
@@ -124,27 +130,149 @@ public void LoadAll()
 
             foreach (var path in filePaths)
             {
-                var plugin = new PluginContext(path);
-
+                if (path.Contains("disabled")) continue;
```
roflmuffin, line 133: I don't think this is super required, as we only match against `{PluginName}/{PluginName}.dll` to load plugins, so the user could move the plugin into a disabled subfolder and it would immediately not match I think.

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
```diff
@@ -124,27 +130,149 @@ public void LoadAll()
...
+                    }
+
+                    break;
+                }
+                case "start":
+                case "load":
+                {
+                    if (info.ArgCount < 2)
+                    {
+                        Utilities.ReplyToCommand(caller, "Valid usage: css_plugins start/load [absolute plugin path] (e.g \"plugins/TestPlugin/TestPlugin.dll\")\n", true);
```
roflmuffin, line 198: Noting that this is actually relative path not absolute, we could also probably consider adding support for both kinds of loading.

```suggestion
                        Utilities.ReplyToCommand(caller, "Valid usage: css_plugins start/load [relative plugin path] (e.g \"plugins/TestPlugin/TestPlugin.dll\")\n", true);
```

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
```diff
@@ -124,27 +130,149 @@ public void LoadAll()
...
+                    {
+                        Utilities.ReplyToCommand(caller, $"  0 plugins loaded.", true);
+                    }
+                    else
+                    {
+                        int pluginCount = 1;
+                        foreach (var plugin in _loadedPlugins)
+                        {
+                            Utilities.ReplyToCommand(caller, $"  [{pluginCount}/{_loadedPlugins.Count()}]: \"{plugin.Name}\" (\"{plugin.Version}\")", true);
+                            pluginCount++;
```
roflmuffin, line 187:
```suggestion
                        int pluginCount = 0;
                        foreach (var plugin in _loadedPlugins)
                        {
                            Utilities.ReplyToCommand(caller, $"  [{++pluginCount}/{_loadedPlugins.Count()}]: \"{plugin.Name}\" (\"{plugin.Version}\")", true);
```

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
```diff
@@ -93,8 +69,38 @@ public void LoadAll()
...
+                    }
+                }
+                _loadedPlugins.Add(plugin);
+                plugin.Load();
+            }
+            catch (Exception e)
+            {
+                _loadedPlugins.Remove(plugin);
+                throw;
+            }
```
roflmuffin, line 99: I think if we just add the plugin to the loaded list _after_ we load, then we can ignore the whole exception delete flow.
```suggestion
            foreach (var existingPlugin in _loadedPlugins)
            {
                if (plugin.Name == existingPlugin.Name)
                {
                    throw new Exception("Plugin is already loaded.");
                }
            }
            plugin.Load();
            _loadedPlugins.Add(plugin);
```

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
roflmuffin: On further thought, I think maybe we should assign an autoincrementing plugin number directly to the plugin when we load it, and then store that value. That way you can call `css_plugins unload 1` to unload a plugin without necessarily having to type the full name.

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
roflmuffin: I think we should also do a happy path attempt here and try to load purely based on a filename, i.e if you type `TestPlugin` it will first try `plugins/TestPlugin/TestPlugin.dll`, and then failing that move onto a path based approach.

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
zonical: Just tried this - this line isn't required.

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
zonical: Gave that a shot in my latest commit.

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
zonical: I've added this in my latest commit by storing an id in `PluginContext`. The command syntax for Id's is `css_plugins unload #<number>`

PR #23, opened 2023-11-04
Modified GiveNamedItem to return created entity pointer

Muinez:


KillStr3aK:
This is needed for sure!

PR #28, opened 2023-11-06
Add GetArgAsPlayers as alternative for FindTarget from sourcemod.

IKiwky:


roflmuffin:
I like this idea, but we should probably separate this out into its own utility function, and then call it from within this `GetArgAsPlayers`, or even just let command handlers call it themselves. Is that something you would be comfortable with? I think we could probably just replicate the `FindTarget` method from sourcemod.

managed/CounterStrikeSharp.API/Utilities.cs
```diff
@@ -103,5 +103,36 @@ public static unsafe string ReadStringUtf8(IntPtr ptr)
...
+            if (pattern.StartsWith("#")) 
+            {
+                if (!int.TryParse(pattern.Substring(1), out var userId)) 
+                    return emptyResult;
+                
+                var playerController = playerEntities.Select(player => player).FirstOrDefault(player => player.UserId == userId);
+                
+                return playerController != null ? new List<CCSPlayerController>() { playerController } : emptyResult;
+            }
+            else if (pattern.StartsWith("@"))
```
roflmuffin, line 120: We can remove this redundant else because of the return in the previous if statement.

managed/CounterStrikeSharp.API/Utilities.cs
```diff
@@ -103,5 +103,36 @@ public static unsafe string ReadStringUtf8(IntPtr ptr)
...
+                var playerController = playerEntities.Select(player => player).FirstOrDefault(player => player.UserId == userId);
+                
+                return playerController != null ? new List<CCSPlayerController>() { playerController } : emptyResult;
+            }
+            else if (pattern.StartsWith("@")) 
+            {
+                switch (pattern) 
+                {
+                    case "@all": return playerEntities;
+                    case "@me": return new List<CCSPlayerController> { player };
```
roflmuffin, line 125: We should check if `player` is null here, and if it is, return the empty object.

managed/CounterStrikeSharp.API/Utilities.cs
```diff
@@ -103,5 +103,36 @@ public static unsafe string ReadStringUtf8(IntPtr ptr)
...
+        public static IEnumerable<CCSPlayerController> FindTarget(CCSPlayerController? player, string pattern) {
+            var emptyResult = Enumerable.Empty<CCSPlayerController>();
+            var playerEntities = Utilities.FindAllEntitiesByDesignerName<CCSPlayerController>("cs_player_controller");
+            
+            if (pattern.StartsWith("#")) 
+            {
+                if (!int.TryParse(pattern.Substring(1), out var userId)) 
+                    return emptyResult;
+                
+                var playerController = playerEntities.Select(player => player).FirstOrDefault(player => player.UserId == userId);
```
roflmuffin, line 116: This select is unnecessary, and we should avoid shadowing the outer `player` variable.
```suggestion
                var playerController = playerEntities.FirstOrDefault(p =>p.UserId == userId);
```

managed/CounterStrikeSharp.API/Utilities.cs
```diff
@@ -103,5 +103,36 @@ public static unsafe string ReadStringUtf8(IntPtr ptr)
...
+            {
+                switch (pattern) 
+                {
+                    case "@all": return playerEntities;
+                    case "@me": return new List<CCSPlayerController> { player };
+                }
+            }
+            else {
+                var playerName = pattern.ToLower();
+                var playerController = playerEntities.Select(player => player).FirstOrDefault(player => player.PlayerName.ToLower().Contains(playerName));
```
roflmuffin, line 130: Same as above.
```suggestion
                var playerController = playerEntities.FirstOrDefault(p => p.PlayerName.ToLower().Contains(playerName));
```

roflmuffin (changes requested):
Thanks, I have left some comments 🙏

PR #29, opened 2023-11-06
Added `HudDestination.Alert` and `CGlowProperty::m_bGlowing`

KillStr3aK:


zonical:
Theoretically: if we pull another schema dump to update server.json, won't that remove the glowing property from this version of the dump - thus stopping the key from being generated? The underlying schema code on the C++ side doesn't rely on this file, and the glowing property will still exist for the class regardless. To solve having to update the server.json file on every update, I would recommend creating a partial class of `CGlowProperty` in the `Core/Model` folder and putting the reference there instead.

KillStr3aK:
> Theoretically: if we pull another schema dump to update server.json, won't that remove the glowing property from this version of the dump - thus stopping the key from being generated? The underlying schema code on the C++ side doesn't rely on this file, and the glowing property will still exist for the class regardless. To solve having to update the server.json file on every update, I would recommend creating a partial class of `CGlowProperty` in the `Core/Model` folder and putting the reference there instead.

You are absolutely right in general, the current dump is missing alot of properties and would be better to rehaul on the schema 'dumper' so we won't miss any property.

As long as we don't do another dump this is fine tho, I'd suggest to work on the dumper before using it again. (not sure if we have the source for that? in that case I'd work on it)

Creating a partial class would require a different class name and mess up inheritance (if we don't add `partial` to CGlowProperty, but it would be removed on schemagen), `CBaseEntity` is partial because it inherits from `BaseEntity` (which doesn't really exists on engine level, it is just wrapped like that) and other classes are derived from `CBaseEntity`, thats why it works like that.

zonical:
[This is the tool that generates files for the source2sdk repository (which hosts schema dumps)](https://github.com/neverlosecc/source2gen).

> Creating a partial class would require a different class name and mess up inheritance

The classes that are generated are already marked as `partial`. If you look at `Core/Model/CBaseEntity.cs` you can see an example of a value (`m_vecAbsVelocity`) being accessed by the schema from outside of the generated classes - so it is possible. I believe it would be best to stick to the same coding style that roflmuffin has established.

KillStr3aK:
> I believe it would be best to stick to the same coding style that roflmuffin has established.
I agree, didn't mean to mess it up.
I also understand what you are talking about, to be honest now I can see that it is indeed `partial`, I was relying on the VS's decompiled assembly where they are not partial classes so it was my bad.

Making changes

zonical:
> I agree, didn't mean to mess it up.
> I also understand what you are talking about, to be honest now I can see that they are indeed `partial`, I was relying on the VS's decompiled assembly where they are not partial classes so it was my bad.

It's all good! You have valid intentions for your changes - I'm just thinking about the long term for the project :)

KillStr3aK:
> [This is the tool that generates files for the source2sdk repository (which hosts schema dumps)](https://github.com/neverlosecc/source2gen).

You sure about that source tho?
Their dump has every properties that we miss currently.

roflmuffin (approved):

roflmuffin:
Yeah the dump definitely needs updating, but it's fine to add to a partial class for now, as if the the dump is updated we can just remove the class and the API remains the same

PR #32, opened 2023-11-06
update docs about libicu / Invariant globalization

Apfelwurm:
Since we tested the Invariant globalization option without noticing any problems, here is the updated docs.


roflmuffin (approved):

roflmuffin:
Thank you!

PR #44, opened 2023-11-08
Basic admin system framework (plus some cleanup)

### LINKED ISSUE #41: Add admin functionality
Since CSSharp is taking off rapidly, and large amounts of plugins are being created, I'd like to suggest a long-term solution here.
A lot of plugins have unique **admins.cfg** files. This will be confusing for the average user when adding multiple plugins.

I'd advise CSSharp to add a base **admins.cfg** file that works similar to how it did in SourceMod: [https://wiki.alliedmods.net/Adding_Admins_(SourceMod)](url)

This way, the average user (and plugin creator) will save a lot of time, both in terms of developing an admin system, but also to manage the plugins.

zonical:
The goal of this PR is to resolve #41 by creating a framework for plugins to check if players are admins and if they have relevant permissions for commands. I was inspired by SourceMod's way of handling admins with a central config file, and it seems that the community would like to keep this approach. Admins are defined in `configs/admins.json` and I've included an example with my `steamid64` as an identifier. The `auth_type` key accepts `steamid64`, `steamid32`, `steamid3`, `steamid2` and `name` for the client name.

~~One difference from SourceMod is the way I've chosen to handle flags. Instead of characters in a string, I've decided to use bitwise flags instead. The `Flags` value in `AdminData` is a `uint`. I will leave it for individual plugin developers create their own enum's for permissions (e.g `KICK = 1 << 1`, `BAN = 1 << 2` and so on...). I have also not ported over the password functionality as I've personally never used it. The code should be more than flexible enough to implement a system like that in the future if desired.~~

~~I would like to extend this functionality in the future to do a permissions check for commands before they are executed, but I will leave that for a future PR.~~

P.S: I've also done some misc code cleanup. I've marked `Utilities.ReplyToCommand` as obsolete as an equivalent function has since been implemented in `CommandInfo`. I've added ` console` argument to `CommandInfo.ReplyToCommand` which will print the message to a valid client console if true. Default value is false. Finally, I've moved `AddCommand` out of `GlobalContext` into it's own Utilities class for other parts of the API to have access to it.


roflmuffin:
I'm sure I am sold on the flag idea, I think it is actually quite difficult for newcomers/non devs to understand how to give a player some permissions. Because we are using json & not limited to VDF, we could probably just make these an array of keys, like `["a", "b", "c"]` etc. with room to provide any string enum key as a value.

Just my opinion as well but I think we should remove the "name" admin type and just force a steamid as a requirement. This way we can just parse all the steamids in the admin list as steamid64 and put them in a nice lookup dictionary keyed by steamid64 and not have to worry about querying player props dynamically and just read from steamid.

zonical:
> I'm not sure I am sold on the flag idea, I think it is actually quite difficult for newcomers/non devs to understand how to give a player some permissions. Because we are using json & not limited to VDF, we could probably just make these an array of keys, like `["a", "b", "c"]` etc. with room to provide any string enum key as a value.

I think we can go even further and just step away from bits and lean more into the array of characters - or heck, even just using whole strings like "kick" and "ban'.

> Just my opinion as well but I think we should remove the "name" admin type and just force a steamid as a requirement. This way we can just parse all the steamids in the admin list as steamid64 and put them in a nice lookup dictionary keyed by steamid64 and not have to worry about querying player props dynamically and just read from steamid.

So do you reckon the key of each AdminData object should the the steamid instead? I'm currently using it as a way to easily tell admin definitions apart since JSON doesn't allow comments unlike VDF.

zonical:
Righto - just made a big change.
I've added a new `PermissionHelper` attribute for command methods. This makes it so we can run permission checks in the background before a command is executed so it's as seamless to integrate as possible. An example of a command using this is:

```csharp
[ConsoleCommand("freeze", "Freezes a client.")]
[PermissionHelper("can_manipulate_players")]
public void OnFreezeCommand(CCSPlayerController? caller, CommandInfo command)
{
...
}
```

Tags are simply defined as an array of strings in the JSON file.

managed/CounterStrikeSharp.API/Core/BasePlugin.cs
```diff
@@ -162,6 +163,16 @@ public void AddCommand(string name, string description, CommandInfo.CommandCallb
 
                 var entity = new CCSPlayerController(NativeAPI.GetEntityFromIndex(i + 1));
                 var command = new CommandInfo(ptr, entity);
+
+                // Do not execute command if we do not have the correct permissions.
+                var methodInfo = handler?.GetMethodInfo();
+                var attr = methodInfo?.GetCustomAttribute(typeof(PermissionHelperAttribute), true) as PermissionHelperAttribute;
```
roflmuffin, line 169: You can use the typesafe version of custom attribute using `.GetCustomAttribute<PermissionHelperAttribute>()?.RequiredPermissions`

managed/CounterStrikeSharp.API/Modules/Admin/AdminManager.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Linq;
+using System.Reflection;
+
+namespace CounterStrikeSharp.API.Modules.Admin
+{
+    public partial class AdminData
+    {
+        [JsonPropertyName("auth_type")] public string AuthType { get; set; }
+        [JsonPropertyName("identity")] public string Identity { get; set; }
+        [JsonPropertyName("flags")] public List<string> Flags { get; set; }
```
roflmuffin, line 19: Does System.Text.Json handle deserialisation to a `HashSet`, if it does this would be a better data structure for this for quick flag lookup.

managed/CounterStrikeSharp.API/Modules/Admin/AdminManager.cs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    public static class AdminManager
+    {
+        private static Dictionary<string, AdminData> _admins = new Dictionary<string, AdminData>();
```
roflmuffin, line 24: We should try and type this as a `SteamID` lookup, then the `GetAdminData` method can just do a simple dictionary lookup. We may need to implement `IEquatable<SteamID>` using the `SteamId64` as the underlying value for this to work though.
```suggestion
        private static Dictionary<SteamID, AdminData> _admins = new Dictionary<SteamID, AdminData>();
```

managed/CounterStrikeSharp.API/Modules/Admin/AdminManager.cs
zonical: Didn't work from my testing unfortunately. Might be something we come back to in the future.

managed/CounterStrikeSharp.API/Modules/Admin/AdminManager.cs
zonical: I tried doing this with `IEquatable` but I wasn't able to get it working. I compromised by changing the key type to `ulong` for SteamID64's. I wouldn't be opposed if you wanted to take a crack at implementing `IEquatable`.

zonical:
I'm going all in. `CommandHelper` is another chunk of code ported from an admin system that I was working in. This provides functionality for argument count checking and providing the user with the valid command usage if they have less than the required arguments. 

I've also included the ability to make commands executable only by the server, only by the client, or by both (default). Note: this isn't implementing client-sided commands, I don't think that's possible - but a situation where this could be used is in a `!bring [target]` command. How could the server console teleport a client to them? This now resolves #48 (sorry @switz !). It might be a breaking change, but I personally haven't seen any examples of `AddStandaloneCommand` (or it's previously named counterpart) outside of the API.

switz:
> This now resolves https://github.com/roflmuffin/CounterStrikeSharp/pull/48 (sorry @switz !).

No need to apologize, only spent about 6 minutes on the PR so if you have something better to replace it, that's all good! Will leave mine open until this gets merged just in case, or feel free to close it.

roflmuffin:
Thank you kindly, I have modified some of the dictionary methods and done a squash commit from the fork

PR #48, opened 2023-11-09
feat: Add serverOnly flag to BasePlugin.AddCommand

switz:
This would be a breaking change to the BasePlugin API, let me know if it needs any changes or fixes.


roflmuffin:
I think we could pretty safely default the value  and make the upgrade more seamless

switz:
Fixed in #44

PR #52, opened 2023-11-10
[Windows] feat: Windows support

laper32:
#1 
Warning: Only tested on Windows, I'm not pretty sure whether these changes will break linux build.
Special thanks to @Ciallo-Ani, which helps us a lot to make a huge progress.


laper32:
Seems that everything is ready. Let's GO!

laper32:
Let's GO!

roflmuffin:
Would you be able to remove some of the extra unnecessary console writelines in the managed side.

laper32:
> Would you be able to remove some of the extra unnecessary console writelines in the managed side.

done

roflmuffin:
Thanks so much for this, I will pull down and test Linux as soon as possible and then look at CI

roflmuffin:
There are still quite a few build errors, mostly int he dotnet_host on linux side, mostly to do with the string types. Did you want me to takeover or are you happy to resolve them. They should show up in CI for the latest commit.

laper32:
just go ahead， since I dont have linux machine right now

laper32:
I think I've compiled all compiler issues, based on by CI

PR #57, opened 2023-11-10
fix: Always send "player_chat" event

ws-cs2:
"!" or "/" chat events should always be fired even if CounterStrikeSharp consumes them

In https://github.com/ws-cs2/cs2-surftimer/ we use the "player_chat" event for chat commands and lot's of servers are using both our metamod plugin & CounterStrikeSharp.

We have our detour on HostSay to fix sending "player_chat" but if a server is running CounterStrikeSharp we won't detour but we need CounterStrikeSharp to send "!" and "/" events.


Spazjuaner:
This is important to allot of mods! Please add asap.

ws-cs2:
I think also sending entindex as userid is wrong too

From my testing:

playerslot = entindex - 1
playerslot = userid & 0xFF

if anything this should be entindex - 1 if userid is not available

ws-cs2:
https://github.com/Source2ZE/CS2Fixes/pull/96/commits/fdd038f9de099668562a5f807df86ea566ba4891

roflmuffin:
Handling this as part of #76

PR #62, opened 2023-11-11
`CoreConfig` implementation on the managed side

KillStr3aK:
I've 'attached' single comments on some part of the code, let me know if this requires modification.
I'll make the unmanaged implementation aswell, and a documentation for this later on

I'm also working on the standardized plugin configuration behind the scenes but that will come with a different pr as I'm not done with that yet.


zonical:
The only suggestion I would make is to turn the comments in `ccnfigs/core.json` into a page on the documentation. Some IDE's might not like comments in JSON, like Visual Studio Code by default.

KillStr3aK:
> Some IDE's might not like comments in JSON, like Visual Studio Code by default.
Yes, VSC requires .jsonc to be selected to avoid error highlighting for the comments

> EDIT: Just saw "and a documentation for this later on" - you're one step ahead!
I'll make a doc page for it when I'm done with the configuration standard, also will remove the comments from the `core.json`

docs/src/content/docs/configuration/core-configuration.md
roflmuffin, line 1: I think this file needs to be moved to `docs/reference` folder for it to be auto-picked up by the contents tree.

docs/src/content/docs/configuration/core-configuration.md
```diff
@@ -0,0 +1,26 @@
...
+
+## FollowCS2ServerGuidelines
+
+Per [CS2 Server Guidelines](https://blog.counter-strike.net/index.php/server_guidelines/), certain plugin
+functionality will trigger all of the game server owner's Game Server Login Tokens
+(GSLTs) to get banned when executed on a Counter-Strike 2 game server.
+
+Enabling this option will block plugins from using functionality that is known to cause this.
+This option only has any effect on CS2. Note that this does NOT guarantee that you cannot
+receive a ban.
```
roflmuffin, line 22: I think this comment is redundant (probably a copy pasta)
```suggestion
Note that this does NOT guarantee that you cannot
receive a ban.
```

managed/CounterStrikeSharp.API/Core/CoreConfig.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// <summary>
+        /// <para>
+        /// Per <see href="http://blog.counter-strike.net/index.php/server_guidelines/"/>, certain plugin
+        /// functionality will trigger all of the game server owner's Game Server Login Tokens
+        /// (GSLTs) to get banned when executed on a Counter-Strike 2 game server.
+        /// </para>
+        ///
+        /// <para>
+        /// Enabling this option will block plugins from using functionality that is known to cause this.
+        /// This option only has any effect on CS2. Note that this does NOT guarantee that you cannot
```
roflmuffin, line 65:
```suggestion
        /// Note that this does NOT guarantee that you cannot
```

docs/src/content/docs/configuration/core-configuration.md
KillStr3aK: Committed suggestion, yes it is indeed a reference from SM

managed/CounterStrikeSharp.API/Core/CoreConfig.cs
```diff
@@ -0,0 +1,118 @@
...
+    public static partial class CoreConfig
+    {
+        private static CoreConfigData _coreConfig = new CoreConfigData();
+
+        static CoreConfig()
+        {
+            CommandUtils.AddStandaloneCommand("css_core_reload", "Reloads the core configuration file.", ReloadCoreConfigCommand);
+        }
+
+        [RequiresPermissions("can_reload_core")]
```
roflmuffin, line 85:
```suggestion
        [RequiresPermissions("@css/config")]
```

docs/src/content/docs/configuration/core-configuration.md
KillStr3aK: I've moved the file to the `reference` folder

PR #67, opened 2023-11-11
feat: Provide configuration standard for plugins

KillStr3aK:
This pr implements an automatic configuration parser and generator for plugins that have configuration files.

Plugins can register their config type like this:

```csharp
/// <summary>
/// Nested type showcase
/// </summary>
public class NestedType
{
    [JsonPropertyName("ULongValue")]
    public ulong ULongValue { get; set; } = 0UL;
}

/// <summary>
/// Random configuration class, must derive from BasePluginConfig or implement IBasePluginConfig
/// </summary>
public class MyConfig : BasePluginConfig
{
    [JsonPropertyName("IntValue")]
    public int IntValue { get; set; } = 0;

    [JsonPropertyName("StringValue")]
    public string StringValue { get; set; } = "localhost";

    [JsonPropertyName("FloatValue")]
    public float FloatValue { get; set; } =  1337.0f;

    [JsonPropertyName("NestedValue")]
    public NestedType NestedValue { get; set; } = new NestedType();
}

/// <summary>
/// Other than deriving from BasePlugin, we can tell the API that we are using the `MyConfig` class as our configuration!
/// </summary>
public class CSSharpTests : BasePlugin, IPluginConfig<MyConfig>
{
    public override string ModuleName => "CSSharp Tests";

    public override string ModuleVersion => "1.0.0.0";

    /// <summary>
    /// Implemented interface member so we can store our instance
    /// </summary>
    public MyConfig Config { get; set; }

    /// <summary>
    /// Called then the `ConfigManager` has parsed the configuration file for this plugin
    /// </summary>
    /// <param name="config">Parsed config instance</param>
    public void OnConfigParsed(MyConfig config)
    {
        // Developers can override this property from BasePluginConfig.Version
        if (config.Version == 1)
        {
            // do specific actions if the configuration version is 1..
        }

        // maybe do sanity checks and validation on the config values
        if (config.StringValue != "localhost")
        {
            throw new Exception($"Invalid value has been set to config value 'StringValue': {config.StringValue}");
        }

        // if everything is good, lets save the instance reference
        Config = config;
    }

    {
        // `OnConfigParsed` is called before `Load` is called so we can use our settings immediately
        if (Config.NestedValue.ULongValue > 0)
        {
            // do the required things if the config value satisfies the condition
        }
    }
}
```

Generated configuration file:

```jsonc
// This configuration was automatically generated by CounterStrikeSharp for plugin 'CSSharpTests', at  2023/11/11 10:23:02
{
  "IntValue": 0,
  "StringValue": "localhost",
  "FloatValue": 1337,
  "NestedValue": {
    "ULongValue": 0
  },
  "ConfigVersion": 1
}
```

Any type that can be serialized by `System.Text.Json` library can be used.

Configuration files are located in `configs/plugins/PLUGIN_NAME/PLUGIN_NAME.json`, if the `PLUGIN_NAME` folder does not exists inside the `plugins` folder it will be automatically generated, if the `PLUGIN_NAME.json` is missing, then that will also be generated.
Developers should explicitly set default values for their properties inside their configuration class (as you can see in the example), as they will be used as default values on generation.

Let me know if it needs further changes, currently the only thing I'm missing is regenerating the configuration without changing the currently set things (so only add the new properties), but it might come in the future, currently this is something that developers can use already!


KillStr3aK:
Ah accidentally created the pr with the changes from the #62

KillStr3aK:
Note: of course the JsonProperty attribute is not a must, just in case you want to set the serialization name for the property, otherwise it will be the property name

zonical:
Absolutely epic stuff from you again Killstreak, ka pai.

KillStr3aK:
Resolved conflicts.

roflmuffin:
Can I be a nitpicker and ask if you can add a basic config to the `TestPlugin`, just because I know a lot of people reference that plugin as an example?

KillStr3aK:
Pretty sure! I'll fulfill these requests soon

roflmuffin (approved):

PR #72, opened 2023-11-12
feat: `CBaseModelEntity_SetModel`

KillStr3aK:
Example usage:

```csharp
public class CSSharpTests : BasePlugin
{
    public override string ModuleName => "CSSharp Tests";

    public override string ModuleVersion => "1.0.0.0";

    public static readonly string ModelPath = "characters\\models\\ctm_heavy\\ctm_heavy.vmdl";

    {
        base.RegisterEventHandler<EventPlayerSpawn>((EventPlayerSpawn @event, GameEventInfo info) =>
        {
            CCSPlayerController player = @event.Userid;

            if (player == null || !player.IsValid || !player.PlayerPawn.IsValid)
                return HookResult.Continue;

            Server.NextFrame((() =>
            {
                Server.PrecacheModel(ModelPath); // note that this is wrong here, but precaching the model is required. (precache should happen on map start probably but this snippet only serves showcase.
                player.PlayerPawn.Value.SetModel(ModelPath);
            }));

            return HookResult.Continue;
        });
    }
    ```


roflmuffin (approved):
Prop Hunt begins now

PR #74, opened 2023-11-12
Admin Manager improvements

### LINKED ISSUE #73: AdminManager improvements
This is just a collection of some ideas that people have been discussing in the Discord to improve the AdminManager API:

- Add groups of permissions to simplify how many flags are required. This can probably be enforced by including a syntax for individual flags `@` and groups of flags `#`
- Remove all arbitrary flags from core e.g `can_execute_css_commands`
- Add wildcard support for RequiresPermissions and PlayerHasPermissions e.g `@css/*`
- Add OR support for RequiresPermissions, this can probably be done with a simple bool argument

zonical:
The goal of this PR is to close #73 by adding some requested features into the AdminManager API and to standardize permission flags.

Planned improvements/features:
- [x]  Add groups of permissions to simplify how many flags are required. This can probably be enforced by including a syntax for individual flags `@` and groups of flags `#`
- [x] Remove all arbitrary flags from core e.g `can_execute_css_commands`
   - Now changed to `@css/generic`
- [ ] Add wildcard support for RequiresPermissions and PlayerHasPermissions e.g @css/*
- [x] Add OR support for RequiresPermissions
- [x] Command overrides
- [x] Add support for immunity values


zonical:
This is my suggestion for loading and handling groups. 
- Groups will be defined in a separate file: `configs/admin_groups.json`. It will contain an array of flags, similar to the current system and probably an immunity value to assign for each member of the group. Players are **not** added to this file.

Example:
```json
{
  "#css/admin": {
    "flags": [
      "@css/generic",
      "@css/permission-1",
      "@css/permission-2"
    ],
    "immunity": 100
  }
}
```

- In `configs/admin.json`, There will be a new array, called `groups`. Players that are tagged with groups will inherit the flags of the group.

Example:
```json
  "Erikj": {
    "identity": "76561197960265731",
    "flags": [
      "@css/custom-flag"
    ],
    "groups": [
      "#css/admin"
    ]
  }
```

When you register permissions in the `RequiresPermissions` attribute, you can pass through a group. When the command is executed, AdminManager will check to see if the caller is part of that group. If so, the command will be passed through.

Let me know if this makes sense or not.

roflmuffin:
The idea seems sound to me

zonical:
cc: @crashzk

I've added support for immunity values, similar to SourceMod. Admins and admin-groups can define immunity values. If an admin is assigned to an admin group, and the group has a higher immunity value, the admin will automatically assume that immunity value. You can compare the immunity between two players with `CanPlayerTarget` - using either player controllers or SteamID.

crashzk:
> cc: @crashzk
> 
> I've added support for immunity values, similar to SourceMod. Admins and admin-groups can define immunity values. If an admin is assigned to an admin group, and the group has a higher immunity value, the admin will automatically assume that immunity value. You can compare the immunity between two players with `CanPlayerTarget` - using either player controllers or SteamID.

First of all thanks for this resource, I can't wait to test everything, thanks again.

Another question would be, do you intend to also implement an override system? A quick explanation would be to change access to plugin permissions more easily, instead of editing the source code of the plugin itself.

Example, Plugin X players with access to flag X would have immunity, but with the overrides system I just add the command for this plugin there and change the flag to have access to it and such.

zonical:
> First of all thanks for this resource, I can't wait to test everything, thanks again.
> 
> Another question would be, do you intend to also implement an override system? A quick explanation would be to change access to plugin permissions more easily, instead of editing the source code of the plugin itself.
> 
> Example, Plugin X players with access to flag X would have immunity, but with the overrides system I just add the command for this plugin there and change the flag to have access to it and such.

I might do this in two ways:
Option 1 could be to simply add a `command_overrides` section to the admin and group definitions, like this for example:
```json
    "command_overrides": {
      "css_plugins": true,
      "css": false
    }
```

Or Option 2 could be to create a new file which is loaded on API initialization with can then change the flags that are checked for when they are executed.

zonical:
New feature: OR permission checking support.
I've implemented a new attribute class, `RequirePermissionsOr`, to keep the code clean and so you can stack AND and OR attributes on top of each other if you really want to (all of them will be checked on command execute).

The following example uses groups, but it should work for user permissions as well:
```csharp
[RequiresPermissionsOr("#css/test-group", "#css/admin")]
public void OnTestCommand(CCSPlayerController? caller, CommandInfo command)
{
...
}

With this update, comes a small breaking change. Any permissions that are not marked with `#` for groups or `@` for user permissions will not be recognized. This is being done so we can enforce a standard syntax.

**Comment by zonical** [comment]
Apologies for the length of time on this PR. I know people have been having some issues with the AdminManager API and I hope to get them resolved ASAP.

Command overrides for admins, admin groups, and commands in general will be my priority for tomorrow. I haven't come up with a solution for the wildcard check yet. I'm thinking for the sake of code clarity, I could make it into a third attribute that can be stacked on top of the others. We'll see.

**Comment by Hackmastr** [comment]
Hello @zonical,

Great work so far on this module, thanks!

I have a suggestion (or perhaps an issue with the group-checking feature).

`[RequiresPermissionsOr("#css/test-group", "#css/admin")]` 

Group checking might be an unnecessary feature, since these are created by server admins, and you can't know their names from the plugin (granted, you could read it, but that's not the ideal mechanism). 

A plugin should only check for a flag(s): 
`[RequiresPermissionsOr("@css/generic")]` 

The manager should check if the target has the required permissions, by checking if the user has it or if they belong to a group that has it.

Thanks again for your work!

**Comment by zonical** [comment]
> `[RequiresPermissionsOr("#css/test-group", "#css/admin")]`
> 
> Group checking might be an unnecessary feature, since these are created by server admins, and you can't know their names from the plugin (granted, you could read it, but that's not the ideal mechanism).
> 
> A plugin should only check for a flag(s): `[RequiresPermissionsOr("@css/generic")]`
> 
> The manager should check if the target has the required permissions, by checking if the user has it or if they belong to a group that has it.
> 
> Thanks again for your work!

Kia ora, thank you for your kind words and feedback :)

I had one person suggest that we could both use groups and user permissions when implementing these checks. I think it makes sense as it simplifies the amount of flags that might need to be written if you have a command with requires 10 flags or something - unlikely but possible. I would prefer to keep it so we don't have to circle back around and re-implement it later on. The best thing about the group system is that it's entirely optional; you can get away with just using user permissions all over the place.

**Comment by crashzk** [comment]
>  > First of all thanks for this resource, I can't wait to test everything, thanks again.
> > 
> > Another question would be, do you intend to also implement an override system? A quick explanation would be to change access to plugin permissions more easily, instead of editing the source code of the plugin itself.
> > 
> > Example, Plugin X players with access to flag X would have immunity, but with the overrides system I just add the command for this plugin there and change the flag to have access to it and such.
> 
> I might do this in two ways:
> Option 1 could be to simply add a `command_overrides` section to the admin and group definitions, like this for example:
> ```json
>     "command_overrides": {
>       "css_plugins": true,
>       "css": false
>     }
> ```
> 
> Or Option 2 could be to create a new file which is loaded on API initialization with can then change the flags that are checked for when they are executed.

Hello, first of all, thank you for your answer and your time on the project, everything is perfect for me.

 Regarding the issue I mentioned, why not have both options?  In my eyes, it would be perfect for both to be implemented and to give two options so that we can configure overrides.

 Anyway, I leave it to your discretion, but what I used most would be the separate file, option 2 that you mentioned.

**Comment by zonical** [comment]
I've just pushed another big commit that adds support for overriding permissions of commands. There are three ways to do this.
1) You can define overrides in the `command_overrides` section for each admin in `configs/admins.json`:
```json
"Erikj": {
    "identity": "76561197960265731",
    "immunity": 100,
    "flags": [
      "@css/custom-flag-1",
      "@css/custom-flag-2"
    ],
    "command_overrides": {
      "css_plugins": true,
      "css": false
    }
 ```
 This means that ErikJ will be allowed to execute `css_plugins` and not `css`, regardless of the flags or groups he is part of.
 
 2) Groups can define command overrides in a similar fashion. Each individual admin will inherit the groups command overrides.
 3) You can override the flags of a command in a new file, `configs/admin_overrides.json`:
 ```json
 {
  "example_command": {
    "flags": [
      "@css/cheats"
    ],
    "enabled": true
  }
}
```

Command overrides can be toggled on and off during runtime with `SetCommandOverideState`, likewise with permission and setting the permissions for a command override.

As I'm writing this, I realized that there's a flaw with the current implementation. If you edit the flags, it will use the same attribute classes that are attached to the methods. I will fix this in the next commit with a new parameter that accepts `all` or `partial`.

zonical:
> As I'm writing this, I realized that there's a flaw with the current implementation. If you edit the flags, it will use the same attribute classes that are attached to the methods. I will fix this in the next commit with a new parameter that accepts all or partial.

I need some help to come up with a design to make this idea work.

At the moment, we can define command callbacks with attributes which store their own permission data (`BaseRequiresPermissions.Permissions`). With the command override feature that I implemented in my last commit, the permissions would be overridden correctly. However, it could be more flexible with developers wanting to state if they want all of the override permissions (the functionality of `RequiresPermissions`) or to only have one or more (`RequiresPermissionsOr`). This system also needs to be stackable i.e. being able to place an AND (`RequirePermissions`) on top of an OR (`RequiresPermissionsOr`) and have it still work.

Part of me wonders if a system like this is too complicated and if we should just default to an AND check for command overrides. Let me know your thoughts.

johnoclockdk:
> > As I'm writing this, I realized that there's a flaw with the current implementation. If you edit the flags, it will use the same attribute classes that are attached to the methods. I will fix this in the next commit with a new parameter that accepts all or partial.
> 
> I need some help to come up with a design to make this idea work.
> 
> At the moment, we can define command callbacks with attributes which store their own permission data (`BaseRequiresPermissions.Permissions`). With the command override feature that I implemented in my last commit, the permissions would be overridden correctly. However, it could be more flexible with developers wanting to state if they want all of the override permissions (the functionality of `RequiresPermissions`) or to only have one or more (`RequiresPermissionsOr`). This system also needs to be stackable i.e. being able to place an AND (`RequirePermissions`) on top of an OR (`RequiresPermissionsOr`) and have it still work.
> 
> Part of me wonders if a system like this is too complicated and if we should just default to an AND check for command overrides. Let me know your thoughts.

Hello maybe make a RequirePermissionsFlex for that not sure if it will be named like that

zonical:
Alright, I think I've finished the API for all of the new features. I'm going to be out for a majority of today and I will need to divert focus away to other obligations, so I would be very grateful if people could test my changes and let me know if there are any major bugs or problems.

[You can get the artifacts from here when the build finishes.](https://github.com/roflmuffin/CounterStrikeSharp/actions/runs/6882171474)

zonical:
I've given everything a quick test and I've not run into any issues. I might leave the wildcard support for another time.

roflmuffin:
Thanks, tested locally and edited some groups and flags and everything appears to be working well

PR #79, opened 2023-11-13
feat: managed coreconfig implementation

### LINKED ISSUE #38: Chat detour
First, thanks for a fantastic project! 💯 

I noticed that the chat detour auto dispatches anything that stars with '/' and '!'. This creates a few problems for us since we have some other infrastructure that needs to know about chats that starts with '!'. 

Also I found that it's not possible to implement a '!pause' command since pause is already a command (leads to seg fault). 

I see a couple of solutions here that would fit nicely for us at least. 

1. Always do m_pHostSay
2. Configuration to turn off chat commands

What do you think?

KillStr3aK:
Implemented `CoreConfig` on the unmanaged part aswell, also the schema limits that comes with `FollowCS2ServerGuidelines` option.

```csharp
NativeAPI.SetSchemaValueByName<int>(weapon.Handle, (int)typeof(int).ToDataType(), "CEconEntity", "m_nFallbackPaintKit", 350);
```
image

resolves #38


CMakeLists.txt
```diff
@@ -86,7 +88,7 @@ SET(SOURCE_FILES
 
 if (LINUX)
     # memoverride.cpp is not usable on CMake Windows, cuz CMake default link libraries (seems) always link ucrt.lib
-    set(SOURCE_FILES 
+    set(SOURCE_FILES
```
KillStr3aK, line 89: I guess this was my IDE?

src/core/cs2_sdk/schema.h
```diff
@@ -54,6 +56,45 @@ inline uint64_t hash_64_fnv1a_const(const char *str, const uint64_t value = val_
 }
 
 namespace schema {
+static std::unordered_set<std::string> CS2BadList({
```
KillStr3aK, line 59: I'm down for a better idea, it was simple like this currently

src/core/utils.h
```diff
@@ -21,6 +21,7 @@ inline std::string GameDirectory() {
 }
 
 inline std::string PluginDirectory() { return GameDirectory() + "/addons/counterstrikesharp"; }
```
KillStr3aK, line 23: I've seen that it is being used in the dotnet host but this is pretty misleading? I think it should lead to the plugins folder and we should have a different function to get the root of cssharp

src/core/cs2_sdk/schema.h
KillStr3aK: maybe a normal `vector` instead of `unordered_set` could do better as that's faster

src/core/utils.h
KillStr3aK: I mean it refers to the metamod plugin directory but in overall this is misleading and we should have a `GetRootFolder` or something like that

src/core/coreconfig.cpp
```diff
@@ -0,0 +1,53 @@
...
+    if (!ifs) {
+        V_snprintf(conf_error, conf_error_size, "CoreConfig file not found.");
+        return false;
+    }
+
+    m_json = json::parse(ifs);
+
+    try {
+        PublicChatTrigger = m_json["PublicChatTrigger"];
+        SilentChatTrigger = m_json["SilentChatTrigger"];
```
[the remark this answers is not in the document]
KillStr3aK: Currently you can do

```json
"PublicChatTrigger": "abcdefgh"
```

and you can use commands with

acss
bcss
ccss
....

src/core/utils.h
KillStr3aK: Done

src/core/coreconfig.cpp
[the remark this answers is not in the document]
roflmuffin: I just think the array syntax is more explicit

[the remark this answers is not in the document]
KillStr3aK: on it

src/core/managers/chat_manager.cpp
```diff
@@ -68,16 +68,17 @@ void DetourHostSay(CBaseEntity* pController, CCommand& args, bool teamonly, int
...
+    bool bCommand = bSilent || globals::coreConfig->IsPublicChatTrigger(args[1], prefix);
 
     if (!bSilent) {
         m_pHostSay(pController, args, teamonly, unk1, unk2);
     }
 
     if (bCommand)
     {
-        char *pszMessage = (char *)(args.ArgS() + 2);
+        char *pszMessage = (char *)(args.ArgS() + prefix->length() + 1);
```
KillStr3aK, line 81: I think this needs further tests as I had no time but it was working fine for me, but had crashes in the early state of the implementation of this part

KillStr3aK: okay it seems pretty fine

KillStr3aK:
I'd say its ready if everything fits.

roflmuffin (approved):

PR #89, opened 2023-11-15
docs: use nuget instead

snorux:


KillStr3aK:
Probably both could stay as an option

snorux:
I can put the nuget option as a tip instead and leave the original. Should be fine?

KillStr3aK:
I think that would be fine yes

snorux:
done

roflmuffin:
Thank you!

Issue #71, opened 2023-11-12
The whole design of CounterStrikeSharp


laper32:
Instead of shared data between app domains, I'd better call it: The whole design of `CounterStrikeSharp`.

You could find that `shared data between app domains` is the subset of `the whole design of 'CounterStrikeSharp'`.

Why we discuss it? Well, this project, `CounterStrikeSharp`, will be renamed to a better name in the future, and it will directly affect the future of the project, and the future of the whole ecosystem.

Any progress will be updated in this issue.

For now, yes, we are now using C# for scripting, but if I remember correctly, one of the aims is: any language what supports the format can be integrated easily and can also share data between different languages.

This requires us to make well-designed architecture. Generally, this MUST lead to a result: We are doing an engine, a tiny Operating System, in fact.

If we want to do it, Unreal, Unity, etc. must be important references, and also S&Box, which is the only game what using C# for scripting in Source 2.

Now, back to our question: How many layers should we have?
1. Engine layer.
    It can also be called Unmanaged layer (aka, not managed by the next layer), which handles such things:
    - Interact with native engine.
    - Provides infrastructures, which may include:
      - Reverse-engineering tools
      - Centralized logging
      - Hacked functions
      - Game-specific information
    - Distributed folder layout
  It should also be easily expanded.
2. Runtime layer.
    It handles such things:
    - Provide user space infrastructures, may include:
      - GC
      - User friendly API
      - Simplified error handling
      - Distributed folder layout
         It inherits the engine layer, and adds user space specific layers.
         It won't contradict the engine layer folder layout.
    - Wrapper of engine layer APIs
    - Application infrastructures.
  Based on the above, in general, we would select something VM-based, like Android of Java, Unity of C#, Unreal of Blueprint.
  Also, this layer should also be expanded easily.
  Our choice is C#.
3. Application layer.
  Since previous layers have already provided all infrastructure what you need, you now can write your own application.
  These applications should be restricted in their own space, and MUST NOT let them pollute any other, even any parent layers.
  
So, we have discussed a lot, what should it look like?
For distributed folder layout:
```
. // addons/counterstrikesharp
|
|---bin                                                1)
|---config                                             2)
|---log                                                3)
|    |---L20231101.log
|    \---L20231102.log
|---gamedata                                           4)
|    |---counterstrikesharp.games.kv
|    \---cs2.games.kv
|---modules                                            5)
|    |---managed                                     5.a)
|    \---unmnaged                                    5.b)  
[...]

PR #91, opened 2023-11-15
Made at least some kind of modularity for plugins

Muinez:
I added `shared` folder containing libraries to be shared by plugins

```
counterstrikesharp
|-- shared
|   |-- Utils
|       |-- Utils.deps.json
|       |-- Utils.dll
|       |-- Utils.pdb
```
both core and module plugin should be dependent on shared library (in this case, `Utils`)

example core plugin:
```
public class TestPlugin : BasePlugin
{
    public override string ModuleName => "TestPlugin";

    public TestPlugin()
    {
        AddApi<ITestApi>(new TestApi());
    }
}
    
public class TestApi : ITestApi
{
    public void PrintSomething()
    {
        Console.WriteLine("12345");
    }
}
```

example module plugin:
```
public class TestModule : BasePlugin
{
    public override string ModuleName => "TestModule";

    {
        var api = GetApi<ITestApi>()!;

        api.PrintSomething();
    }
}
```

example shared library:
```
public interface ITestApi
{
    public void PrintSomething();
}
```

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/API.cs
+++ b/managed/CounterStrikeSharp.API/Core/API.cs
@@ -124,12 +124,23 @@ namespace CounterStrikeSharp.API.Core
 			}
 		}
 
-        public static void IssueClientCommand(int clientindex, string command){
+        public static void IssueClientCommand(int slot, string command){
 			lock (ScriptContext.GlobalScriptContext.Lock) {
 			ScriptContext.GlobalScriptContext.Reset();
-			ScriptContext.GlobalScriptContext.Push(clientindex);
+			ScriptContext.GlobalScriptContext.Push(slot);
 			ScriptContext.GlobalScriptContext.Push(command);
 			ScriptContext.GlobalScriptContext.SetIdentifier(0xCA5BA982);
+			ScriptContext.GlobalScriptContext.Invoke();
+			ScriptContext.GlobalScriptContext.CheckErrors();
+			}
+		}
+
+        public static void IssueClientCommandFromServer(int slot, string command){
+			lock (ScriptContext.GlobalScriptContext.Lock) {
+			ScriptContext.GlobalScriptContext.Reset();
+			ScriptContext.GlobalScriptContext.Push(slot);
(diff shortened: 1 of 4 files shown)


KillStr3aK:
this is under discuss in https://github.com/roflmuffin/CounterStrikeSharp/issues/71

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
```diff
@@ -19,27 +19,71 @@
...
+                }
+                catch
+                {
+                    var assembly = _sharedAssemblies.First(a => a.name.Name == name.Name);
+                    return assembly.assembly;
+                }
+            };
+
+            LoadSharedLibs();
+        }
```
KillStr3aK, line 67: `McMaster` has its own implementation of shared libraries so I'd say this is needless

laper32:
please see #71

managed/CounterStrikeSharp.API/Core/GlobalContext.cs
Muinez: > `McMaster` has its own implementation of shared libraries so I'd say this is needless

what exactly are you talking about?

frederikstonge: > > `McMaster` has its own implementation of shared libraries so I'd say this is needless
> 
> what exactly are you talking about?

https://www.nuget.org/packages/McMaster.NETCore.Plugins

They use this package to load plugin assemblies.

partiusfabaa:
I don't think it can be done any better. You have to add

KillStr3aK:
> I don't think it can be done any better. You have to add

We already have "plans" about the implementation

partiusfabaa:
> > I don't think it can be done any better. You have to add
> 
> We already have "plans" about the implementation

I think its already a good option, so why do anything else if its already there?

roflmuffin:
Hi @Muinez, would you be able to review #253 and see if this solves the use case for this PR?

PR #95, opened 2023-11-15
EventPlayerChat - Use the result of the palyer_chat event, consume the event if we should

charliethomson:


roflmuffin:
I think this may be better achieved by adding a command listener to the `say` command and returning a Handled/Stop when you want to reject.

charliethomson:
the say command listener also doesn't seem to consume the message, haven't dug to find out why, minimal example: 
```cs
AddCommandListener("say", OnPlayerChat);

// ..

public HookResult OnPlayerChat(CCSPlayerController player, CommandInfo info)
{
    Logging.LogTrace("OnPlayerChat");

    Console.WriteLine($"PlayerSay: {info.GetArg(1)}");

    return HookResult.Handled; // or Stop
}
```
image
image

PR #105, opened 2023-11-20
feat: Added ability to GiveNamedItem using the new CsItem Enum

LordFetznschaedl:
I have created an enum of all items I have found online with the item string. 
In addition the possibility was added for players to GiveNamedItem using the Enum.

Example for M4A4:
player.GiveNamedItem(CsItem.M4A1);


managed/CounterStrikeSharp.API/Modules/Utils/EnumUtils.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CounterStrikeSharp.API.Modules.Utils
+{
+    public static class EnumUtils
+    {
+        public static string GetEnumMemberAttributeValue<T>(T enumValue)
```
KillStr3aK, line 12: should be nullable if it can return null?

KillStr3aK (commented):
I dont think this is a needed change in general, would be handy for ppl but maybe make the enum available in a library

roflmuffin (approved):

managed/CounterStrikeSharp.API/Modules/Utils/EnumUtils.cs
LordFetznschaedl: String is a referenced type and thus can be already null. Good practice to have checks anyways before using them further. But yes, I have changed it to nullable string.

managed/CounterStrikeSharp.API/Modules/Utils/CsItem.cs
```diff
@@ -0,0 +1,244 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CounterStrikeSharp.API.Modules.Utils
+{
+    public enum CsItem
```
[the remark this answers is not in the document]
LordFetznschaedl: The list comes from the internet. Just have forgotten to remove the defuser. this is done with the new commit

roflmuffin:
The enum looks good, but could we move and namespace this under `Modules/Entities/Constants`, similar to ItemDefinitionIndex?

LordFetznschaedl:
> The enum looks good, but could we move and namespace this under `Modules/Entities/Constants`, similar to ItemDefinitionIndex?

Change has been made.

roflmuffin:
Legendary, thank you sir

Issue #109, opened 2023-11-21
Respawn / Spawn


Mbk10201:
Hi,

I was wondering if the respawn / spawn method has been implemented or not yet.

Btw i love the API, its much better than traditional pawn.

KillStr3aK:
No implementation yet but its being taking care of

PR #114, opened 2023-11-22
Add methods to respawn players

KillStr3aK:
yes, both call needed for respawn


KillStr3aK:
resolves #109

Mbk10201:
Thankss that was fast <3

PR #121, opened 2023-11-23
Feature: ProcessTargetString() & GetPlayerFromSteamId()

CharlesBarone:
Primary goal of this pull request is to add the target string functionality previously available in SourceMod to CounterStrikeSharp.

For reference to the related SM feature: https://sm.alliedmods.net/new-api/commandfilters/ProcessTargetString & https://wiki.alliedmods.net/Admin_commands_(sourcemod)


KillStr3aK:
Implementation seems good to me, however instead of a new 'module' I think it should be included in the 'Commands' module

roflmuffin:
We could probably also add a helper or extension method to CommandInfo to treat arg as a target string

Hackmastr:
Any chance we could add @aim?
:D

roflmuffin:
> Any chance we could add `@aim`? :D

We'll probably need trace rays working first :(

Hackmastr:
> > Any chance we could add `@aim`? :D
> 
> We'll probably need trace rays working first :(

Thought so. Unfortunate.

roflmuffin:
Thank you for the contribution, I have made a couple changes after testing (using playerpawn for life state) & removing the steamid2 escaping (I think we should just make clients input a quoted string, i.e. `"#STEAM_0:0:1"`). 

This is probably pretty opinionated, but I have removed the TargetCount & TargetAcquisitionException as I don't find they add much value to the consumer, ultimately the result will always be an array of players given the context of a target string, so we should just let that be handled on the plugin side. Since `TargetResult` is an `IEnumerable` anyway, the caller can use existing LINQ methods like `Single` if they want to throw when there are more than 1 players coming back from the target string.

PR #129, opened 2023-11-25
Implement Core & Plugin Service Collection

roflmuffin:
- Refactor for some of the core bootstrapping & application startup logic to use a service collection
- Adds a new `IPluginServiceCollection<T>` type which is scanned for which allows plugins to provide their own `ConfigureServices` method to extend & utilise dependency injection.
- Plugin types are resolved using a container rather than with an `Activator.CreateInstance` which allows plugin authors to inject dependencies into their root plugin class.

```csharp
public class TestInjectedClass
{
    private readonly ILogger<TestInjectedClass> _logger;

    public TestInjectedClass(ILogger<TestInjectedClass> logger)
    {
        _logger = logger;
    }

    public void Hello()
    {
        _logger.LogInformation("Hello World from Test Injected Class");
    }
}

public class TestPluginServiceCollection : IPluginServiceCollection<SamplePlugin>
{
    public void ConfigureServices(IServiceCollection serviceCollection)
    {
        serviceCollection.AddScoped<TestInjectedClass>();
    }
}

public class SamplePlugin : BasePlugin
{
    private readonly TestInjectedClass _testInjectedClass;
    public SamplePlugin(TestInjectedClass testInjectedClass)
    {
        _testInjectedClass = testInjectedClass;
    }

    {
        _testInjectedClass.Hello();
    }
}
```


laper32:
looks great, the structure looks very nice than before.

managed/CounterStrikeSharp.API/Core/Hosting/IScriptHostConfiguration.cs
```diff
@@ -7,25 +7,25 @@
...
     /// <summary>
-    /// Gets the absolute path to the directory that contains all CounterStrikeSharp files.
+    /// Gets the absolute path to the directory that contains CounterStrikeSharp files.
     /// e.g. /game/csgo/addons/counterstrikesharp
     /// </summary>
     string RootPath { get; }
     
     /// <summary>
-    /// Gets the absolute path to the directory that contains all CounterStrikeSharp plugins.
+    /// Gets the absolute path to the directory that contains ounterStrikeSharp plugins.
```
KillStr3aK: typo

managed/CounterStrikeSharp.API/Core/CoreConfig.cs
```diff
@@ -78,49 +82,58 @@ public static partial class CoreConfig
         public static bool FollowCS2ServerGuidelines => _coreConfig.FollowCS2ServerGuidelines;
     }
 
-    public static partial class CoreConfig
```
KillStr3aK, line 81: are you sure that `CoreConfig` should be retrieved as a service instead of `static` access?

managed/CounterStrikeSharp.API/Core/GameData.cs
```diff
@@ -31,33 +32,45 @@ public class Offsets
     [JsonPropertyName("linux")] public int Linux { get; set; }
 }
 
-public static class GameData
```
KillStr3aK, line 34: are you sure that `GameData` should be retrieved as a service instead of `static` access?

KillStr3aK:
this is fantastic and amazing structure in general, way more better than how it is currently and I pretty much like every change.
Curious how it will turn out in the end, great work!!

KillStr3aK:
I assume in the end it will resolve our discussion in #71

managed/CounterStrikeSharp.API/Core/GameData.cs
roflmuffin: GameData is still a static class, GameDataProvider is the service used from core to load the game data. Ideally I would prefer if everything pulled configs from the container (even plugin configs), because it is just better in every way than statics (i.e. we have context of which plugin called it through container etc.)

managed/CounterStrikeSharp.API/Core/CoreConfig.cs
roflmuffin: I am happy to leave it as a service for now, since this is only internal for now and we haven't used it anywhere statically. Ideally anything that reads coreconfig will have access to this through DI

KillStr3aK: we haven't directly (other than the schema field limits) but plugins could access it using `CoreConfig.<option>` without the requirement of DI and services

managed/CounterStrikeSharp.API/Core/GameData.cs
KillStr3aK: I agree on DI overall, just making sure we want to use services for gamedata and coreconfig (also other things that are not mentioned if there is any)

managed/CounterStrikeSharp.API/Core/CoreConfig.cs
roflmuffin: Yeah and that should still work since it has static members with are initialised with the Load method called by core; ideally this class has the same treatment as the game data provider, where we have a static version (like `CoreConfig` is now) but also an instance version which can be injected

Issue #26, opened 2023-11-05
System.Reflection.TargetInvocationException thrown when accessing fields of `@event.Userid` in a GameEventHandler


miguno:
The server throws an exception when trying to use the simple [GampleEventHandler from the docs](https://docs.cssharp.dev/features/game-events/). The only change is that I used `Console.Writeln()` instead of the `Log()` method in the example. The exception is thrown when the `@event` is being read.

```csharp
[GameEventHandler]
public HookResult OnPlayerConnect(EventPlayerConnect @event, GameEventInfo info) {
  // This works:
  Server.PrintToChatAll($"Player {@event.Name} has connected!");

  // But this will trigger an exception:
  Server.PrintToChatAll($"{@event.Userid.IsBot} will trigger an exception!");

  // @event.Userid will give you a reference to a CCSPlayerController class.
  // However, accessing any of its fields will throw an exception!
  HookResult.Continue;
}
```

Exception:

```
Creating NetChan for user 0 'm0NESY' @ 192.168.0.44:59053
Associating NetChan       m0NESY    192.168.0.44:59053[1] (192.168.0.44:59053) wit
CServerSideClientBase::Connect( name='m0NESY', userid=0, fake=0, chan->addr=192.16
Client 0 'm0NESY' signon state SIGNONSTATE_NONE -> SIGNONSTATE_CONNECTED
Server waking up from hibernation
L 11/05/2023 - 19:50:43: "m0NESY<0><[U:1:81210237]><>" connected, address "192.168
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at System.Runtime.InteropServices.Marshal.ReadByte(IntPtr ptr, Int32 ofs)
   at CounterStrikeSharp.API.Core.ScriptContext.GetResult(Type type, Byte* ptr) in /__w/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Core/Script
Context.cs:line 407
   at CounterStrikeSharp.API.Core.ScriptContext.GetResultHelper(fxScriptContext* context, Type type) in /__w/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrike
Sharp.API/Core/ScriptContext.cs:line 391
[...]

PR #133, opened 2023-11-25
Check if userid is valid before accessing its fields, and explain why

miguno:
This fixes the main example in the current documentation that showcases how to work with game events.  Right now, that code example will consistently trigger runtime exceptions.

This PR will fix the code example and also educates the user reading the documentation why that is required.

Fixes https://github.com/roflmuffin/CounterStrikeSharp/issues/26.


miguno:
Fixes https://github.com/roflmuffin/CounterStrikeSharp/issues/26

Xellement (approved):

PR #142, opened 2023-11-27
Entity Handle Overhaul

roflmuffin:
- Creates a new type `NativeEntity` which extends `NativeObject`.
  - This type holds a `CEntityHandle` (`CHandle<EntityInstance>`) in memory, created either from the raw uint passed in, or from a valid pointer passed in.
- Subsequent existing references to `Handle` now pull the pointer using "EntitySystem" on managed side, which is a port of the entity system code from hl2sdk. This lets us do handle -> entity lookups quickly without paying the cost of invoking a native.
- `IsValid` now checks if A: the handle is valid and B: that the entity pointer derived from the handle is also valid
- Marks `EntityIndex` obsolete (with fail build flag on) so we can avoid the pain of `EntityIndex.Value.Value` that was previously happening, and instead added a `.Index` uint32 property directly to `CEntityInstance`. 
- Adds a new `As<T>` helper that can be used to create a new instance of another native object from the current one; this will prevent usages of the constructor when we simply want to cast, i.e. `CPlayer_WeaponServices` to `CCSPlayer_WeaponServices` can be done by using `weaponServices.As<CCSPlayer_WeaponServices>()`
- Adds `.Slot` property to `CCSPlayerController` which is the player slot (index - 1)

Overall I'm hoping this PR prevents unnecessary crashes by adding more protection to entity accessing and `IsValid` checks.


KillStr3aK:
This seems 🔥, neat stuff!!!!

managed/CounterStrikeSharp.API/NativeEntity.cs
```diff
@@ -0,0 +1,21 @@
+using System.Runtime.InteropServices;
+using CounterStrikeSharp.API.Modules.Entities;
+using CounterStrikeSharp.API.Modules.Utils;
+
+namespace CounterStrikeSharp.API;
+
+public abstract class NativeEntity : NativeObject
+{
+    public new IntPtr Handle => EntitySystem.GetEntityByHandle(EntityHandle) ?? 0;
```
KillStr3aK, line 9: what about using `IntPtr.Zero` instead of 0? its the same just it got my attention

roflmuffin: I just got lazy haha

src/scripting/natives/natives_entities.cpp
```diff
@@ -64,12 +110,20 @@ CEntityIdentity* GetFirstActiveEntity(ScriptContext& script_context) {
...
 REGISTER_NATIVES(entities, {
     ScriptEngine::RegisterNativeHandler("GET_ENTITY_FROM_INDEX", GetEntityFromIndex);
     ScriptEngine::RegisterNativeHandler("GET_USERID_FROM_INDEX", GetUserIdFromIndex);
     ScriptEngine::RegisterNativeHandler("GET_DESIGNER_NAME", GetDesignerName);
     ScriptEngine::RegisterNativeHandler("GET_ENTITY_POINTER_FROM_HANDLE",
                                         GetEntityPointerFromHandle);
+    ScriptEngine::RegisterNativeHandler("GET_ENTITY_POINTER_FROM_REF", GetEntityPointerFromRef);
+    ScriptEngine::RegisterNativeHandler("GET_REF_FROM_ENTITY_POINTER", GetRefFromEntityPointer);
+    ScriptEngine::RegisterNativeHandler("GET_CONCRETE_ENTITY_LIST_POINTER", GetConcreteEntityListPointer);
+    ScriptEngine::RegisterNativeHandler("IS_REF_VALID_ENTITY", IsRefValidEntity);
```
roflmuffin, line 126: Though we aren't using these anymore, I'm still keeping them around in case I ever need to debug the entity system on managed side

managed/CounterStrikeSharp.API/NativeEntity.cs
roflmuffin: Fixed it now.

roflmuffin:
> its not an important thing but could you also update the `TestPlugin` `MinimumApiVersion` to a higher one? as its using dynhooks and other stuffs, for e.g. the new entity handle things that cannot run on artifact v33 pretty sure haha

I've just set it to latest.

PR #152, opened 2023-11-29
Improve plugin setup docs

switz:
This comes after a conversation in the discord in which we identified that some files here are not necessary to upload as they are already included via CSS.

Screenshot 2023-11-29 at 15 48 59@2x

https://discord.com/channels/1160907911501991946/1160907912445710480/1179520129709641900

cc: @KillStr3aK


roflmuffin (approved):
Thank you!

PR #160, opened 2023-12-01
[DRAFT] Implementing memory patches

KillStr3aK:
util related things and `CMemPatch` is borrowed from cs2fixes

this is a pretty wanky implementation, I think we should drop the gameconfig dependency from `CMemPatch` as thats why I made it like how it is currently.

assume we have the following gamedata:

```json
{
    "ServerMovementUnlock": {
      "signatures": {
        "library": "server",
        "windows": "\\x76\\x2A\\xF2\\x0F\\x10\\x57\\x2A\\xF3\\x0F\\x10\\x2A\\x2A\\x0F\\x28\\xCA\\xF3\\x0F\\x59\\xC0",
        "linux": "\\x0F\\x87\\x2A\\x2A\\x2A\\x2A\\x49\\x8B\\x7C\\x24\\x2A\\xE8\\x2A\\x2A\\x2A\\x2A\\x66\\x0F\\xEF\\xED\\x66\\x0F\\xD6\\x85\\x2A\\x2A\\x2A\\x2A"
      },

      // we are still inside 'ServerMovementUnlock' so the patch would be for the signature above.
      "patches": {
        "windows": "\\xEB",
        "linux": "\\x90\\x90\\x90\\x90\\x90\\x90"
      }
    }
}
```

```csharp
// this is a local build so no logger instance as I have not installed transitive packages
public class SamplePlugin : BasePlugin
{
    public override string ModuleName => "Sample Plugin";

    private MemoryPatch patch = new MemoryPatch("ServerMovementUnlock", "ServerMovementUnlock");

    {
        if (patch.PerformPatch())
        {
            Console.WriteLine("Performed patch {0} ({1})", patch.Name, patch.Signature);
        }
        else
        {
            Console.WriteLine("Unable to perform patch {0} ({1})", patch.Name, patch.Signature);
        }
    }

    public override void Unload(bool hotReload)
    {
        patch.UndoPatch();
    }
}
```

hotreloading the plugin will undo the patch, then perform it again:

image

also patches performed using the wrapper method is automatically reverted on unload:

```csharp
public class SamplePlugin : BasePlugin
{
    public override string ModuleName => "Sample Plugin";

    {
        CreateMemoryPatch("ServerMovementUnlock", "ServerMovementUnlock");
    }
}
```

image

> [!CAUTION]
> Plugins that create their patches using the wrapper classes only and call `PerformPatch` are responsible for reverting the patch on unload. Creating patches with `CreateMemoryPatch` will be automatically reverted.


src/scripting/natives/natives_memory.cpp
```diff
@@ -159,6 +162,26 @@ void* GetNetworkVectorElementAt(ScriptContext& script_context)
     return &vec->Element(index);
 }
 
+bool CreateMemoryPatch(ScriptContext& scriptContext)
+{
+    auto pszSignatureName = scriptContext.GetArgument<const char*>(0);
+    auto pszName = scriptContext.GetArgument<const char*>(1);
```
KillStr3aK, line 168: wrong order

KillStr3aK:
This will require change or rehaul in the future for sure, but its something we can use.
Since the core does not require any patch I've disabled auto performing them on load to give full control to the plugins, this can be changed later on if we'd need any patch to be performed internally

Oylneechan:
I know this is already closed, but I did test on my own server. it seem like the signature and patches can't be in other gamedata  json file, except the main gamedata.json file

KillStr3aK:
this will be recreated in managed

PR #165, opened 2023-12-02
Replace documenation with docfx

busheezy:
This will replace the current documentation with docfx so the API can be automatically generated alongside the documentation.

If you have a large version of the logo I would love to use it on the landing page.

An instance of the gh pages is at https://busheezy.github.io/CounterStrikeSharp/index.html

Thanks for everything.


pedrotski:
There are too many buttons on the home page, and it doesn't really elevator pitch a first time visitor. It should just have 1 call to action saying "Get Started Now!" that brings them straight to the installation guide.

switz:
Can we use an svg logo instead of a png?

roflmuffin:
Thank you for the contribution!

Issue #125, opened 2023-11-23
[Bug?] Admin @css/root


ChocOranger:
Using plugin CS2Rcon it requires the admin flag @css/rcon. Setting that flag for my admin works as expected however using @css/root doesn't grant admin access to the plugin.
Is @css/root bugged or am I missing something in how it is used?

The plugin code looks correct to me
https://github.com/LordFetznschaedl/CS2Rcon/blob/be0cef22665d7e6f5b4882dc02d48b79ce84d046/CS2Rcon/CS2Rcon.cs#L26

roflmuffin:
The permissions don't work in the same way as source mod, the `root` does not imply all of the other permissions; admins will need to explicitly have the `@css/rcon` flag or be in a group with `@css/rcon` to enable use of the command.

crashzk:
@roflmuffin but wouldn't it be possible to say that if you have access to @css/rcon you would have access to all flags, even without configuring it for the admin?

Or could this be a problem?

ChocOranger:
The documentation states
@css/root # Magically enables all flags and ignores immunity values. (Same wording used in Sourcemod)

What is @css/root intended purpose?

adammyllykoski:
My two cents here would be to simply change the behavior, and let the `@css/root` flag's intended purpose be to act as a "superadmin". If I add a player as a "root" admin, I would expect them to get access to all commands, which in my world should include both @css flags and perhaps also all custom flags added by plugin creators. This would also match the behavior people are used to from SourceMod. 

As of right now, this flag doesn't make any sense, IMO. 

Managing flags for each and every plugin, for every admin, may become a time-consuming task over time when adding new plugins that require new flags. Let's remember that many just want to have plugins to enhance their experience with their friends/teammates, and not necessarily use it for public community servers where fine-tuned control may be needed.

ChocOranger:
I fully agree with @adammyllykoski and @crashzk (although I think he meant @css/root) that it should get access to all flags. People are already misunderstanding its purpose.
The best place to add a master admin is in the framework then all users get access rather than everyone having to create their own.

I do however feel it should be all flags not just @css ones.

roflmuffin:
@zonical Do you have thoughts on whether we should make @css/root a wildcard flag?

zonical:
I have two alternatives:
1) add a new key to `admins.json` which specifies an all-command override
2) turn the `@css/root` flag into `@root`, so it's not linked to any plugin or system

[...]

PR #168, opened 2023-12-03
Changed the command accessibility for some built-in administrator commands

ManifestManah:
Changed the administrator permission flags required in order to use some of the built-in commands in order to allow users with the root flag access to also be able to use the built-in commands explicitly available to users with the cvar, generic and config access flags. To make it align with the documentation https://docs.cssharp.dev/admin-framework/defining-admins/ in which it is stated that root access users have all other flags enabled, albeit seemingly root access users couldn't use the css_plugins command.


roflmuffin:
I think we are tracking this in #125 and plan on using star wildcards per namespace, i.e. `@css/*` to track flag permissions.

ManifestManah:
Ah alright, I'll go ahead and close this issue then :-)

PR #170, opened 2023-12-03
Prevent calling natives on non-main thread

Poggicek:
Developers could accidentally call natives on another thread which could be a real hell to debug. This pull requests add a check to InvokeNative to make sure that the current thread id matches the server thread id, otherwise it will throw an exception to managed.

### DIFF
--- a/src/core/globals.cpp
+++ b/src/core/globals.cpp
@@ -78,6 +78,7 @@ ChatManager chatManager;
 ServerManager serverManager;
 
 GetLegacyGameEventListener_t* GetLegacyGameEventListener = nullptr;
+std::thread::id gameThreadId;
 
 void Initialize()
 {
--- a/src/core/globals.h
+++ b/src/core/globals.h
@@ -107,6 +107,7 @@ extern CGameConfig* gameConfig;
 typedef IGameEventListener2 *GetLegacyGameEventListener_t(CPlayerSlot slot);
 
 extern GetLegacyGameEventListener_t* GetLegacyGameEventListener;
+extern std::thread::id gameThreadId;
 
 void Initialize();
 // Should only be called within the active game loop (i e map should be loaded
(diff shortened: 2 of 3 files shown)


tk1114632:
Hi, my plugin used to call HTTP request in another thread. But after this update, because of the preventing of calling natives on non-main thread, the Async HTTP request no long works. I'm quite new to the project and C#, is there way to work around? Thank you in advance.

w7rus:
Had same issue, found a solution.

You should make a service that holds a blocking queue of Func<T>/Actions to process them on a main thread. After enqueuing a work item, it returns a task from TaskCompletionSource to await the result in other threads. You can register that service as singleton.

Service:
```csharp
public class NativeFunctionCallQueueService : INativeFunctionCallQueueService
{
    private readonly BlockingCollection<Action> _queue = new();
    
    public Task<T> Enqueue<T>(Func<T> func, CancellationToken cancellationToken = default)
    {
        var tcs = new TaskCompletionSource<T>();
        
        cancellationToken.Register(() => tcs.TrySetCanceled());
        
        _queue.Add(() =>
        {
            try
            {
                tcs.TrySetResult(func());
            }
            catch (Exception ex)
            {
                tcs.TrySetException(ex);
            }
        }, cancellationToken);
        
        return tcs.Task;
    }

    public Task Enqueue(Action action, CancellationToken cancellationToken = default)
    {
        var tcs = new TaskCompletionSource();
        
        cancellationToken.Register(() => tcs.TrySetCanceled());
        
        _queue.Add(() =>
        {
            try
            {
                action.Invoke();
                
                tcs.TrySetResult();
            }
            catch (Exception ex)
            {
                tcs.TrySetException(ex);
            }
        }, cancellationToken);
        
        return tcs.Task;
    }

    public void ProcessQueue()
    {
        foreach (var action in _queue.GetConsumingEnumerable()) action?.Invoke();
    }
}
```

Processing of work items:
```csharp
    {
        RegisterListener<Listeners.OnServerPreWorldUpdate>((e) =>
        {
            _nativeFunctionCallQueueService.ProcessQueue();
        });

        base.Load(hotReload);
    }
```

Usage:
```csharp
...

if (await _nativeFunctionCallQueueService.Enqueue(() => !Server.IsMapValid(Mapname), cancellationToken))
    throw new MapnameInvalidMapException(Mapname);
                
...
```

PR #173, opened 2023-12-05
docs: win32 related

laper32:
Some people asking me why unable to startup, and when I'm digging in, I found that many of them running on server side, and generally, server side won't install vc redist by default.

This pr provides the doc to inform ppl using windows to install it.


crashzk:
@laper32 I believe the documentation will change to `docfx`: https://github.com/roflmuffin/CounterStrikeSharp/pull/165

It would be better to request this addition in his PR now and then it will be merged together.

PR #174, opened 2023-12-05
Add Entity Output Hooks

roflmuffin:
Continuation of work from #167. Adds pre & post hook modes, as well as the ability to provide wildcard hooks using `*` for both class name & output name.


roflmuffin:
> Request before merge: can we have some docs on this feature, please? I don't mean for this to come off as rude. Still, there have been many features merged recently without relevant documentation to explain how to use it, and it's not particularly user-friendly to read the TestPlugin when there is so much content in it.

I added an example plugin with this feature iteration at least, you can find it in the `examples` subfolder of this working tree.

PR #176, opened 2023-12-06
Update docs to use releases and add download buttom

johnoclockdk:


johnoclockdk:
> We can remove the comment about being signed in to GitHub if we link to releases I think

sure thing

Issue #182, opened 2023-12-07
Admin rights not working


asapverneri:
I'm having an issue with my cs2 windows dedicated server where i'm using daffyyyy's CS2-SimpleAdmin plugin.
I reached out to daffyyyy and we made sure that it's not plugin issue.
The problem is that i don't have permission to use admin rights. All commands works throught game server's console tho.
However typing css_admins_list to server's console it shows that i'm an admin.
I've tried using group instead of flags and Steamid64 as identity, but those changes doesn't make any difference.

I'm currently running latest cssharp (v112) and my admins.json looks like this:
```
{
  "yay": {
    "identity": "STEAM_0:0:XXXXXXXX",
	"immunity": 100,
	"flags": [
      "@css/reservation",
      "@css/generic",
      "@css/kick",
      "@css/ban",
      "@css/unban",
      "@css/vip",
      "@css/slay",
      "@css/changemap",
      "@css/cvar",
      "@css/config",
      "@css/chat",
      "@css/vote",
      "@css/password",
      "@css/rcon",
      "@css/cheats",
      "@css/root",
      "@resetscore/admin"
    ]
  }
}
```

KillStr3aK:
probably worth waiting for #143 to be merged and ensure that the issue is still persists

roflmuffin:
Please try again on build 115 if you are using Windows.

asapverneri:
Works with v115. Problem solved i guess :)

Issue #64, opened 2023-11-11
Can I fork out a Chinese version of your warehouse? I originally planned to use it myself, but I want to keep it on Github


Max66888:
Can I fork out a Chinese version of your warehouse? I originally planned to use it myself, but I want to keep it on Github

https://github.com/Max66888/MatchZy-Chinese

KillStr3aK:
Please keep these discussions on the related plugin's repository as it is not related to CounterStrikeSharp.

Max66888:
> [贡献者 ](https://github.com/roflmuffin/CounterStrikeSharp/commits?author=KillStr3aK)
> ###

Sorry, I just realized that I sent the wrong message

PR #185, opened 2023-12-07
Fix SteamId on Windows Server #182

TheR00st3r:
On Windows `NativeAPI.GetPlayerAuthorizedSteamid(Slot);` returns the SteamId32. 
SteamID checks during constructions if the value is a valid SteamId64. If not the Base value will be added.

This fixes the not working admin rights on windows systems #64.


TheR00st3r:
The fix is not really related to Windows but the bug only occures on Windows. 

On Linux CS2 Server `GetPlayerAuthorizedSteamid` return a SteamId64. On Windows it returns a SteamId32.

roflmuffin:
> The fix is not really related to Windows but the bug only occures on Windows.
> 
> On Linux CS2 Server `GetPlayerAuthorizedSteamid` return a SteamId64. On Windows it returns a SteamId32.

That is wild

PR #194, opened 2023-12-11
feat: Added support for custom gamedata files.

CharlesBarone:
This pull request aims to add custom gamedata functionality to CounterStrikeSharp.

Currently you can only add custom offsets/signatures by editing the included gamedata.json file. This change adds the ability to make additional json files in the gamedata directory and have them loaded.

This is beneficial when you want to add offsets/signatures that are not default to CSSharp and want to avoid having to update gamedata.json every time you update CSSharp...

Proof of functionality:
image

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/GameData.cs
+++ b/managed/CounterStrikeSharp.API/Core/GameData.cs
@@ -53,7 +53,7 @@ public sealed class GameDataProvider : IStartupService
             foreach (string filePath in Directory.EnumerateFiles(_gameDataDirectoryPath, "*.json"))
             {
                 string jsonContent = File.ReadAllText(filePath);
-                Dictionary<string,LoadedGameData> loadedMethods = JsonSerializer.Deserialize<Dictionary<string, LoadedGameData>>(jsonContent)!;
+                Dictionary<string, LoadedGameData> loadedMethods = JsonSerializer.Deserialize<Dictionary<string, LoadedGameData>>(jsonContent)!;
 
                 foreach (KeyValuePair<string, LoadedGameData> loadedMethod in loadedMethods)
                 {
@@ -68,6 +68,10 @@ public sealed class GameDataProvider : IStartupService
                 if (loadedMethods != null)
                 {
                     _logger.LogInformation("Successfully loaded {Count} game data entries from {Path}", loadedMethods.Count, filePath);
+                }
+                else
+                {
+                    _logger.LogWarning("Unable to load game data entries from {Path}, game data file is empty", filePath);
                 }
             }
         }
--- a/managed/CounterStrikeSharp.API/Core/GameData.cs
+++ b/managed/CounterStrikeSharp.API/Core/GameData.cs
@@ -34,34 +34,54 @@ public class Offsets
 
 public sealed class GameDataProvider : IStartupService
 {
-    private readonly string _gameDataFilePath;
+    private readonly string _gameDataDirectoryPath;
     public Dictionary<string,LoadedGameData> Methods;
     private readonly ILogger<GameDataProvider> _logger;
 
     public GameDataProvider(IScriptHostConfiguration scriptHostConfiguration, ILogger<GameDataProvider> logger)
     {
         _logger = logger;
-        _gameDataFilePath = Path.Join(scriptHostConfiguration.GameDataPath, "gamedata.json");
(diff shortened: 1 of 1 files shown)


managed/CounterStrikeSharp.API/Core/GameData.cs
```diff
@@ -34,34 +34,54 @@ public class Offsets
...
+                Dictionary<string,LoadedGameData> loadedMethods = JsonSerializer.Deserialize<Dictionary<string, LoadedGameData>>(jsonContent)!;
+
+                foreach (KeyValuePair<string, LoadedGameData> loadedMethod in loadedMethods)
+                {
+                    if (Methods.ContainsKey(loadedMethod.Key))
+                    {
+                        _logger.LogWarning("GameData Method \"{Key}\" loaded a duplicate entry from {filePath}.", loadedMethod.Key, filePath);
+                    }
+                    
+                    Methods[loadedMethod.Key] = loadedMethod.Value;
```
KillStr3aK, line 65: loading order is undefined and if two plugins are using the same key and expect them to be different it will cause interfere between them

@roflmuffin I'm not saying that this (^) is something that should be taken care of, but I'd say we could switch to a loading implementation that returns the loaded gamedata instance and it could be stored in the plugins so they could also get the offsets from the given instance instead of the current way (I'm good with both tho)

managed/CounterStrikeSharp.API/Core/GameData.cs
```diff
@@ -34,34 +34,54 @@ public class Offsets
...
     {
         try
         {
-            Methods = JsonSerializer.Deserialize<Dictionary<string, LoadedGameData>>(File.ReadAllText(_gameDataFilePath))!;
+            Methods = new Dictionary<string, LoadedGameData>();
+
+            foreach (string filePath in Directory.EnumerateFiles(_gameDataDirectoryPath, "*.json"))
+            {
+                string jsonContent = File.ReadAllText(filePath);
+                Dictionary<string,LoadedGameData> loadedMethods = JsonSerializer.Deserialize<Dictionary<string, LoadedGameData>>(jsonContent)!;
```
KillStr3aK, line 56:
```diff
- Dictionary<string,LoadedGameData> loadedMethods = JsonSerializer.Deserialize<Dictionary<string, LoadedGameData>>(jsonContent)!;
+ Dictionary<string, LoadedGameData> loadedMethods = JsonSerializer.Deserialize<Dictionary<string, LoadedGameData>>(jsonContent)!;
```

managed/CounterStrikeSharp.API/Core/GameData.cs
```diff
@@ -34,34 +34,54 @@ public class Offsets
...
+                        _logger.LogWarning("GameData Method \"{Key}\" loaded a duplicate entry from {filePath}.", loadedMethod.Key, filePath);
+                    }
+                    
+                    Methods[loadedMethod.Key] = loadedMethod.Value;
+                }
+                
+                if (loadedMethods != null)
+                {
+                    _logger.LogInformation("Successfully loaded {Count} game data entries from {Path}", loadedMethods.Count, filePath);
+                }
```
KillStr3aK, line 71: maybe some nice log about empty gamedata file?

managed/CounterStrikeSharp.API/Core/GameData.cs
CharlesBarone: I mean I think at that point you'd cause errors in both plugins if they don't match. While they might also be the same offsets/sigs (For example dev A needs the same non-CSSharp native offset and gets it from plugin B's repo).

If they don't match and are the same key, then at that point you have incompatibility between the 2 plugins and one of them will have to change their key.

Either you'd have to accept this limitation or change the format these are stored in.

managed/CounterStrikeSharp.API/Core/GameData.cs
CharlesBarone: See commit.

managed/CounterStrikeSharp.API/Core/GameData.cs
CharlesBarone: See commit.

managed/CounterStrikeSharp.API/Core/GameData.cs
roflmuffin: I agree with Charles, the keys should be unique enough to be shared across plugins

roflmuffin (approved):
LGTM

PR #211, opened 2023-12-18
Fix typo in getting started guide

B3none:


roflmuffin:
This link has been removed so should no longer be a relevant update.

PR #214, opened 2023-12-20
Added a parameter for people to optionally remove the entity when calling RemoveItemByDesignerName

B3none:
I was caught out by doing `player.RemoveItemByDesignerName("weapon_c4");`, it removed the item from the players inventory but didn't remove the entity.

Here's a PR to allow people to specify whether or not they would like to also remove the entity.

New usage:
`player.RemoveItemByDesignerName("weapon_c4", true);`

### DIFF
--- a/managed/CounterStrikeSharp.API/Utilities.cs
+++ b/managed/CounterStrikeSharp.API/Utilities.cs
@@ -90,7 +90,7 @@ namespace CounterStrikeSharp.API
                 item = weapon;
             }
             
-            if(item != null && item.Value != null)
+            if (item != null && item.Value != null)
             {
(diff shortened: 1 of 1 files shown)


roflmuffin:
Would it be possible to declare this as an overload rather than an optional parameter, as adding an optional parameter is a breaking API change due to how optional parameters are handled at compile time.

B3none:
> Would it be possible to declare this as an overload rather than an optional parameter, as adding an optional parameter is a breaking API change due to how optional parameters are handled at compile time.

Ah I wasn't aware of this, have updated the PR accordingly 👍

roflmuffin (approved):

PR #216, opened 2023-12-22
Change TerroristsPlanned to TerroristsPlanted in RoundEndReason

Ravid-A:
I think the reason is to indicate that the terrorists planted the bomb and it exploded no Planned the explosion


roflmuffin:
I think you will need to keep the existing one to preserve compatibility (and add an obsolete attribute), and then make a correct version, as I am trying to reduce breaking changes as much as possible at the moment

B3none:
I've addressed the PR comments, so this is good to go 👍

B3none (approved):

PR #220, opened 2023-12-24
Fix getting started image

pedrotski:
The previous relative path didn't work, as it pointed to:

https://docs.cssharp.dev/docs/guides/images/gameinfogi-example.png

### DIFF
--- a/docfx/docs/guides/getting-started.md
+++ b/docfx/docs/guides/getting-started.md
@@ -20,7 +20,7 @@ If you're more of a visual person, here is a <a href="https://www.youtube.com/wa
 3. Create a new line underneath `Game_LowViolence    csgo_lv` and add `Game    csgo/addons/metamod`.
 4. Restart your game server.
 
-Your `gameinfo.gi` should look like <a href="images/gameinfogi-example.png" target="_blank">this</a>. Type `meta list` in your server console to see if Metamod is loaded.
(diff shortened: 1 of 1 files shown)


docfx/docs/guides/getting-started.md
```diff
@@ -20,7 +20,7 @@ If you're more of a visual person, here is a <a href="https://www.youtube.com/wa
 3. Create a new line underneath `Game_LowViolence    csgo_lv` and add `Game    csgo/addons/metamod`.
 4. Restart your game server.
 
-Your `gameinfo.gi` should look like <a href="images/gameinfogi-example.png" target="_blank">this</a>. Type `meta list` in your server console to see if Metamod is loaded.
+Your `gameinfo.gi` should look like <a href="https://docs.cssharp.dev/images/gameinfogi-example.png" target="_blank">this</a>. Type `meta list` in your server console to see if Metamod is loaded.
```
roflmuffin, line 23: Does this work if you just prefix it with / instead of the full url?

pedrotski: ****

> Does this work if you just prefix it with / instead of the full url?

I don't know, sorry. I just followed how the logo was done when I tried before. I don't have a local test environment for it either, so I thought I would just send it with the full path as it's 100% going to work.

pedrotski: Actually it may need ../../

PR #221, opened 2023-12-24
New menu implementation (Like FACEIT)

Stimayk:
The implementation of the menu through chat is not ideal in my opinion, so I saw how it is implemented in Faceit aand decided to redo the old menu. The control is also done via chat, but displayed in hud via PrintToCenterHTML.
I know about the problems with using PrintToCenterHTML and tried to solve them with OnTick, which I think worked out well in the end.
My code is not perfect and has some crutches, but it works perfectly. Edits and improvements are welcome.
I tested on several plugins and none of them required me to update the way the existing menus work.
So this is a great solution in terms of not having to rewrite most of the existing plugins later on.
Basically only changed the display method, and added improvements to how page flipping works, added menu closing and other things that such a menu display should have.

What I got:
image

Like Faceit does:
image


chycken:
finally the question of the menu came up. It is very confusing that the choice is made via chat. I don't know much about it, but isn't it possible to create something like the radio menu? So that you don't absolutely have to use the chat for the selection, just the numbers on the keyboard (similar to csgo)

zonical:
> Maybe im wrong about something, but i think to remove the chat menu is so self, maybe make a separate class? For example, HintMenu or CenterMenu?

I agree. This should be split into a series of separate classes instead of removing chat menus altogether. I like using chat menus and removing them altogether would be quite saddening.

CHR15cs:
I implemented menus via PrintToCenterHTML for my project as well.
Something I also added was a flag to the menu constructor to not close the menu after a selection of an option, instead it needs to be closed using a specified option. This could be helpfull for menus where you might want to select more than one option

roflmuffin:
We should extract the menu type and interface from each other so plugin authors can write the definitions of menu items once, but use whichever implementation they want (chat, center text, console etc.)

Stimayk:
I've taken all of the above into account and added the ability to choose to display the menu, now the menu can be displayed in all 4 available locations (Chat, Center, CenterHTML and Console)

Here is an example of how it is done in the code:
Chat
`var chatmenu = new ChatMenu($"ChatMenu", DisplayMethod.Chat);`
Center
`var centermenu = new ChatMenu($"CenterMenu", DisplayMethod.Center);`
CenterHTML
`var centerhmtlmenu = new ChatMenu($"CenterHtmlMenu", DisplayMethod.CenterHtml);`
Console
`var consolemenu = new ChatMenu($"ConsoleMenu", DisplayMethod.Console);`

But the code probably still needs tweaking, although everything works as well

managed/CounterStrikeSharp.API/Modules/Menu/ChatMenu.cs
```diff
@@ -1,164 +1,312 @@
...
+        public ChatMenuOption AddMenuOption(string display, Action<CCSPlayerController, ChatMenuOption> onSelect, bool disabled = false)
+        {
+            var option = new ChatMenuOption(display, disabled, onSelect);
+            MenuOptions.Add(option);
+            return option;
+        }
     }
 
-    public ChatMenuOption AddMenuOption(string display, Action<CCSPlayerController, ChatMenuOption> onSelect, bool disabled = false)
+    public class ChatMenuInstance : BasePlugin
```
frederikstonge, line 50: Is this supposed to be a BasePlugin now?

frederikstonge: It would be better to pass a `BasePlugin`instance to the method `OpenMenu` in ChatMenus :

```csharp
public static class ChatMenus
{
    public static void OpenMenu(BasePlugin plugin, CCSPlayerController player, ChatMenu menu)
    {
        ActiveMenus[player.Handle] = new ChatMenuInstance(plugin, player, menu);
        ActiveMenus[player.Handle].Display();
    }
}
```

This way, you can use the BasePlugin to register your listener in the constructor and avoid the menu being a plugin itself.

Stimayk: Thank you very much for the improvement, updated the code

frederikstonge: You still need to update ChatMenus static class. Search in your IDE for the class.

B3none:
Looks like the changes that you've made to this PR are not quite what the guys are after. I plan to re work what you have done into a the style they're expecting tonight as it's a good change, just implemented in slightly the wrong way.

roflmuffin:
> Looks like the changes that you've made to this PR are not quite what the guys are after. I plan to re work what you have done into a the style they're expecting tonight as it's a really good change, just implemented in slightly the wrong way.

To this end, it would be best to extract the menu implementation to something (like `IMenu`), and then create two implementations, CenterMenu and ChatMenu (the old implementation), and users can swap out to the new implementation if they wish. The only problem is that CenterMenu requires a tick based display system vs. chat requires only one call to Display() to work properly and fitting that into the abstraction may prove difficult.

B3none:
I'm closing this as I've branched off of these changes for my PR which updates the menu management system.

https://github.com/roflmuffin/CounterStrikeSharp/pull/268

PR #241, opened 2024-01-06
[DNM] Added initial cross plugin communication functionality (via event handlers)

B3none:
**This has NOT been tested yet**
This is more of a brain dump / proof of concept (I think) PR.

## Things to consider
- How to handle plugin event name clashes
- This means that all plugins can see / access all other plugins

# Example: **without data**
Event sending example (plugin 1):
```c#
SendPluginEvent("RetakesSpawnsComplete");
```

Event handling example (plugin 2):
```c#
[PluginEventHandler]
public void OnRetakesSpawnsComplete()
{
    // Do something now retakes spawns are complete.
}
```

# Example: **with data**
Event sending example (plugin 1):
```c#
SendPluginEvent("RetakesBombsiteSelected", new object[] { "A" });
```

Event handling example (plugin 2):
```c#
[PluginEventHandler]
public void OnRetakesSpawnsComplete(object[] data)
{
    // Do something with the bombsite data.
}
```

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Application.cs
+++ b/managed/CounterStrikeSharp.API/Core/Application.cs
@@ -82,6 +82,7 @@ namespace CounterStrikeSharp.API.Core
             AdminManager.AddCommands();
 
             _pluginManager.Load();
+            _pluginManager.PopulateLoadedPlugins();
 
             for (var i = 1; i <= 9; i++)
             {
--- a/managed/CounterStrikeSharp.API/Core/Attributes/Registration/PluginEventHandlerAttribute.cs
+++ b/managed/CounterStrikeSharp.API/Core/Attributes/Registration/PluginEventHandlerAttribute.cs
@@ -1,0 +1,9 @@
+using System;
+
+namespace CounterStrikeSharp.API.Core.Attributes.Registration;
+
+[AttributeUsage(AttributeTargets.Method)]
+public class PluginEventHandlerAttribute : Attribute
+{
+    
+}
--- a/managed/CounterStrikeSharp.API/Core/BasePlugin.cs
+++ b/managed/CounterStrikeSharp.API/Core/BasePlugin.cs
@@ -21,6 +21,7 @@ using System.Linq;
 using System.Reflection;
 using CounterStrikeSharp.API.Core.Attributes;
 using CounterStrikeSharp.API.Core.Attributes.Registration;
+using CounterStrikeSharp.API.Core.Plugin;
 using CounterStrikeSharp.API.Core.Translations;
 using CounterStrikeSharp.API.Modules.Admin;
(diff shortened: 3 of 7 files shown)


KillStr3aK:
`object[] data` might be a bad decision, I've also been working on something similar to this but haven't touched it for like 3 weeks here: [custom-listeners](https://github.com/roflmuffin/CounterStrikeSharp/compare/main...KillStr3aK:CounterStrikeSharp:feature/custom-listeners)

I had plan to rework it to drop the delegate methods as they could break easily and use a somewhat closer implementation to what FiveM does, maybe you could get on it as I'm currently unable to continue working on it

B3none:
> `object[] data` might be a bad decision, I've also been working on something similar to this but haven't touched it for like 3 weeks here: [custom-listeners](https://github.com/roflmuffin/CounterStrikeSharp/compare/main...KillStr3aK:CounterStrikeSharp:feature/custom-listeners)
> 
> I had plan to rework it to drop the delegate methods as they could break easily and use a somewhat closer implementation to what FiveM does, maybe you could get on it as I'm currently unable to continue working on it
> 
> image image image
> 
> you could also use the non lambda version aswell
> 
> ## Also IPC is still on the table so sharing types and data will be implemented anyway, but the idea of custom listeners / events (aka sp forwards) could come handy

I definitely think that solution looks nicer. Doing it your way works but they won't be able to share listener types during development so it'd all be guess work for a plugin developer. It would allow for checking / validation during the hook call to ensure that they're using the correct schema for the plugin version then throw a warning if it looks like they're mismatched so that could also work.

KillStr3aK:
> > `object[] data` might be a bad decision, I've also been working on something similar to this but haven't touched it for like 3 weeks here: [custom-listeners](https://github.com/roflmuffin/CounterStrikeSharp/compare/main...KillStr3aK:CounterStrikeSharp:feature/custom-listeners)
> > I had plan to rework it to drop the delegate methods as they could break easily and use a somewhat closer implementation to what FiveM does, maybe you could get on it as I'm currently unable to continue working on it
> > image image image
> > you could also use the non lambda version aswell
> > ## Also IPC is still on the table so sharing types and data will be implemented anyway, but the idea of custom listeners / events (aka sp forwards) could come handy
> 
> I definitely think that solution looks nicer. Doing it your way works but they won't be able to share listener types during development so it'd all be guess work for a plugin developer. It would allow for checking / validation during the plugin event call to ensure that they're using the correct schema for the plugin version which is firing the plugin event; then it allows us to throw a warning if it looks like they're mismatched so that could also work.

not exactly, listener definition should be declared in the shared type implementation and the plugin should reference from that, however the delegate is just not the way we should do this in general, I'd be happy for a fivem like implementation, but in that case the guessing part will be there, however I don't consider that as a problem as long as the developers has their own docs about their shared "events"

PR #245, opened 2024-01-08
Add DarkRed to ChatColors class to follow naming convention

B3none:
I kept the original value too since people may already be using it


roflmuffin:
Are you able to add an obsolete attribute to the old darkred member?

PR #251, opened 2024-01-09
Add player value getter

partiusfabaa:
Allows you to get a player's value, such as their store balance or their SteamId. This is useful if the plugin has message settings in the config, for example for ads or hud.

## Plugin 1:
```cs
using CounterStrikeSharp.API;
using CounterStrikeSharp.API.Core;

namespace TestPlugin;

public class TestPlugin : BasePlugin
{
    public override string ModuleName { get; }
    public override string ModuleVersion { get; }

    private readonly int[] _balance = new int[65];

    private readonly PlayerValuesManager _valuesManager;

    public TestPlugin(PlayerValuesManager valuesManager)
    {
        _valuesManager = valuesManager;
    }

    public override void Load(bool hotReload)
    {
        _valuesManager.AddValueGetter("player_balance", i => $"{_balance[i]}");
        _valuesManager.AddValueGetter("player_name", i => $"{Utilities.GetPlayerFromSlot(i).PlayerName}");
        _valuesManager.AddValueGetter("player_steamid", i => $"{new SteamID(Utilities.GetPlayerFromSlot(i).SteamID)}");
        
        RegisterListener<Listeners.OnClientConnected>(slot => _balance[slot] = Random.Shared.Next());
        RegisterListener<Listeners.OnClientDisconnectPost>(slot => _balance[slot] = -1);
    }
}
```

## Plugin 2:
```cs
using CounterStrikeSharp.API;
using CounterStrikeSharp.API.Core;

namespace TestPlugin2;

public class TestPlugin2 : BasePlugin
{
    public override string ModuleName { get; }
    public override string ModuleVersion { get; }
    
    private const string Input = "Player balance: {player_balance}\nPlayer name: {player_name}";

    private readonly PlayerValuesManager _valuesManager;

    public TestPlugin2(PlayerValuesManager valuesManager)
    {
        _valuesManager = valuesManager;
    }

    public override void Load(bool hotReload)
    {
        AddCommand("css_myinfo", "", (player, info) =>
        {
            if (player == null) return;
            
            player.PrintToCenter(_valuesManager.ReplacePlayerValues((int)player.Index - 1, Input));
        });
		
	AddCommand("css_mysteamid", "", (player, info) =>
        {
            if (player == null) return;
            
            player.PrintToChat(_valuesManager.GetValue((int)player.Index - 1, "player_steamid"));
        });
    }
}
```

PR #253, opened 2024-01-10
Shared Plugin APIs/Capabilities

roflmuffin:
Adds a basic shared API & capability system, which can be used for inter plugin communication, i.e. plugins can expose a contract library (placed into the `shared/` subfolder), and then can expose these at a player or plugin level. A common use case would be a "balance/currency" plugin that provides a storage mechanism for users VIP currency, and then other plugins can spend or increase this through some sort of shared balance service.

```csharp
    public static PlayerCapability<IBalanceHandler> BalanceCapability { get; } = new("myplugin:balance");
```
```csharp
        Capabilities.RegisterPlayerCapability(BalanceCapability, player => new BalanceHandler(player));
```

```csharp
public class BalanceHandler : IBalanceHandler
{
    private readonly CCSPlayerController _player;

    // This could be a database, a file, or a dictionary like this.
    internal static readonly Dictionary<CCSPlayerController, decimal> Balances = new();

    public BalanceHandler(CCSPlayerController player)
    {
        _player = player;
    }

    public decimal Balance
    {
        get => Balances.TryGetValue(_player, out var balance) ? balance : 0;
        set => Balances[_player] = value;
    }

    public decimal Add(decimal amount)
    {
        return Balance += amount;
    }

    public decimal Subtract(decimal amount)
    {
        return Balance -= amount;
    }
}
```


Muinez:
All plugins are loaded one by one, but what if a plugin wants to register itself as a module to another plugin?  The module will not be able to get the library from another plugin when loading because it is not sure that it was loaded after the plugin that provides the API. Then in each module will have to boilerplate Server.NextFrame or AddTimer in Load to get the library already in it. 
Correct me if I am wrong.

Also about https://github.com/roflmuffin/CounterStrikeSharp/pull/251 

It was made for another purpose, but having modularity now it can be done without making changes in CSS. Fabius did it to be able to get player values without changing the code, so that server administrators can add messages in the config depending on the plugins they have, for example there is a rpg plugin and it allows you to get rpg_player_credits, and the server administrator wanted to say in the advertising plugin to display his credits to the player in a message, he adds a message like: `Your rpg credits: {rpg_player_credits}`.

roflmuffin:
> All plugins are loaded one by one, but what if a plugin wants to register itself as a module to another plugin? The module will not be able to get the library from another plugin when loading because it is not sure that it was loaded after the plugin that provides the API. Then in each module will have to boilerplate Server.NextFrame or AddTimer in Load to get the library already in it. Correct me if I am wrong.
> 
> Also about #251
> 
> It was made for another purpose, but having modularity now it can be done without making changes in CSS. Fabius did it to be able to get player values without changing the code, so that server administrators can add messages in the config depending on the plugins they have, for example there is a rpg plugin and it allows you to get rpg_player_credits, and the server administrator wanted to say in the advertising plugin to display his credits to the player in a message, he adds a message like: `Your rpg credits: {rpg_player_credits}`.

Anything that requires the presence of another plugin should be done in OnAllPluginsLoaded.

B3none:
Do you think it's worth making the including part of the shared types nicer?

I think that including the reference to the project like this assumes that everyone has the plugins / directory structure for dev setup in the same way. I think this will become a minefield if a plugin contains references to multiple shared libs and we will have to start including bash / ps1 scripts to clone different repos in the intended directory structure.

https://github.com/roflmuffin/CounterStrikeSharp/pull/253/files#diff-027f70acabc5f8267a59be4b05682ed8163f1ab8cd7eebe855e74994e3ee2e24R11

roflmuffin:
> Do you think it's worth making the including part of the shared types nicer?
> 
> I think that including the reference to the project like this assumes that everyone has the plugins / directory structure for dev setup in the same way. I think this will become a minefield if a plugin contains references to multiple shared libs and we will have to start including bash / ps1 scripts to clone different repos in the intended directory structure for plugin development.
> 
> https://github.com/roflmuffin/CounterStrikeSharp/pull/253/files#diff-027f70acabc5f8267a59be4b05682ed8163f1ab8cd7eebe855e74994e3ee2e24R11

Well ideally you would publish your "api" package to nuget, and each plugin can just reference the nuget package. It does feel like we will need to start declaring some sort of plugin manifest file though because the dependency tree will get a bit hard to track.

B3none:
I've tested this locally now with cross-plugin communication. It took a bit of trial and error to figure out exactly how it works for plugin development, but it's actually a pretty nice solution and seems to solve all of the use cases for plugin communication I can think of 👍

Muinez:
still no OnAllPluginsLoaded

roflmuffin:
> still no OnAllPluginsLoaded

Two minutes too early 😉

managed/CounterStrikeSharp.API/Core/Capabilities/PlayerCapability.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections.Generic;
+
+namespace CounterStrikeSharp.API.Core.Capabilities;
+
+public sealed class PlayerCapability<T> where T : class
+{
+    public string Name { get; }
+    internal static readonly Dictionary<string, List<Func<CEntityInstance, T?>>> Providers = new();
```
Muinez, line 8: why CEntityInstance and not CBasePlayerController?

roflmuffin: Allows the implementation to further be extended later down the line to accept other entity types, so eventually you could have `RegisterPawnCapability` for instance, or any generic `RegisterCapability<T>` where `T : CEntityInstance`. In that way plugin authors could attach any extra data they want to a `CPlantedC4` for instance.

This is all hidden away from the end user at this point as we are supporting purely player & plugin at this stage.

roflmuffin: I suppose in this case it is a sealed `PlayerCapability` class so it should probably be changed to `CCSPlayerController`

managed/CounterStrikeSharp.API/Core/Capabilities/Capabilities.cs
```diff
@@ -0,0 +1,33 @@
...
+        Func<CCSPlayerController, T> supplier) where T : class
+    {
+        if (!PlayerCapability<T>.Providers.ContainsKey(capability.Name))
+        {
+            PlayerCapability<T>.Providers.Add(capability.Name, new());
+        }
+
+        PlayerCapability<T>.Providers[capability.Name].Add(entity =>
+        {
+            if (entity is CCSPlayerController player)
```
Muinez, line 25: then I think this check is unnecessary

roflmuffin: Fixed https://github.com/roflmuffin/CounterStrikeSharp/pull/253/commits/5deaa5337a11823011a77751cb088cf3d9164941

Muinez (approved):

roflmuffin:
I have widened the types, to allow for basic getters using primitive types, which should lessen the burden on consumers to have to compile against a given contract, i.e. a plugin can declare a `Decimal` or a `float` type which might store some balance, and its just up to the provider to provide a decimal value for this player. If they want to interact with the plugin in some other way (i.e. call methods), they will still need an API contract.

`public static PlayerCapability<Decimal> BalanceCapabilityDecimal { get; } = new("myplugin:balance_decimal");`

Muinez:
I think it's a good idea

It would also be cool if `PluginCapability` and `PlayerCapability` contained a plugin and when calling `Get` also passed it and that it was possible to subscribe to plugin unloading. In this case, the core plugin, if another plugin registers itself as a module, could automatically unregister it when it is unloaded

Issue #260, opened 2024-01-14
Linux Build 146 segfault


M1kep:
After reinstalling the .NET framework to work around #258, the server segfaults on startup:

Backtrace:
```
#0  0x00007fffc43aae1f in ?? () from /home/steam/game/game/csgo/addons/counterstrikesharp/dotnet/shared/Microsoft.NETCore.App/8.0.1/libcoreclr.so
#1  0x00007fffc43aa878 in coreclr_initialize () from /home/steam/game/game/csgo/addons/counterstrikesharp/dotnet/shared/Microsoft.NETCore.App/8.0.1/libcoreclr.so
#2  0x00007fffdb87c0a5 in ?? () from /home/steam/game/game/csgo/addons/counterstrikesharp/dotnet/shared/Microsoft.NETCore.App/8.0.1/libhostpolicy.so
#3  0x00007fffdb8971ee in ?? () from /home/steam/game/game/csgo/addons/counterstrikesharp/dotnet/shared/Microsoft.NETCore.App/8.0.1/libhostpolicy.so
#4  0x00007fffdb8d6e98 in ?? () from /home/steam/game/game/csgo/addons/counterstrikesharp/dotnet/host/fxr/8.0.1/libhostfxr.so
#5  0x00007fffdb8d1c44 in hostfxr_get_runtime_delegate () from /home/steam/game/game/csgo/addons/counterstrikesharp/dotnet/host/fxr/8.0.1/libhostfxr.so
#6  0x00007fffc4abbee7 in CDotNetManager::Initialize() () from /home/steam/game/game/csgo/addons/counterstrikesharp/bin/linuxsteamrt64/counterstrikesharp.so
#7  0x00007fffc4a9c36f in counterstrikesharp::CounterStrikeSharpMMPlugin::Load(int, SourceMM::ISmmAPI*, char*, unsigned long, bool) ()
   from /home/steam/game/game/csgo/addons/counterstrikesharp/bin/linuxsteamrt64/counterstrikesharp.so
#8  0x00007fffc4ea9f2f in CPluginManager::_Load(char const*, int, char*, unsigned long) () from /home/steam/game/game/csgo/addons/metamod/bin/linuxsteamrt64/metamod.2.cs2.so
[...]

PR #261, opened 2024-01-15
.NET8 Upgrade

roflmuffin:
.NET8 upgrade, which currently segfaults on Linux.

- [x] Windows (tested ✔️) 
- [x] Linux (see #260)


frederikstonge:
There's a file with references to net7.0 still. It's a suppression xml file, so nothing serious, but you might want to take a look : 
`managed/CounterStrikeSharp.API/CompatibilitySuppressions.xml`.

Also, you might want to update nuget packages starting with Microsoft.Extensions.* to version 8.0.0 to match framework version : 
`Microsoft.Extensions.`
`Microsoft.Extensions.`
`Microsoft.Extensions.`

B3none:
Tested this again, and it's working as expected for Linux 👍

roflmuffin:
> There's a file with references to net7.0 still. It's a suppression xml file, so nothing serious, but you might want to take a look : `managed/CounterStrikeSharp.API/CompatibilitySuppressions.xml`.
> 
> Also, in the `CounterStrikeSharp.API` project, you might want to update nuget packages starting with Microsoft.Extensions.* to version 8.0.0 to match framework version : `Microsoft.Extensions.Hosting` `Microsoft.Extensions.Hosting.Abstractions` `Microsoft.Extensions.Localization.Abstractions` `Microsoft.Extensions.Logging` `Serilog.Extensions.Logging`
> 
> and potentially : `Microsoft.DotNet.ApiCompat.Task`

I am somewhat hesitant to do this, mainly because we can still run existing plugins built for .NET7 in the new .NET8 runtime, but if we upgrade the abstractions packages etc. to .NET8, then this may break existing plugins that are using `ILogger` (as an example) from a .NET7 reference package. I would say the best approach may be to offer some grace period and then upgrade these packages internally.

PR #265, opened 2024-01-17
Log exception if plugin load fails using the `load` command

wiesendaniel:
I've realized this might be missing.
Caught me as i was mostly hot-reloading my plugins during development.


roflmuffin (approved):

B3none:
Going to run the build on this, and if it passes (which I suspect it will) I'll merge it.

B3none:
> Going to run the build on this, and if it passes (which I suspect it will) I'll merge it.

I was going to do this but I didn't realise every push to `main` makes a release. I'll have to wait for @roflmuffin to advise.

wiesendaniel:
I ran into a lot of situation like this in the past. Maybe you want to create a `next` or `release` branch. Merge everything on this and do a single merge to `main`. Let me know if you need me to change the target branch or something.

PR #268, opened 2024-01-18
Menu system updates

B3none:
Updated the menu system to allow for different menu instances.

Before:
```c#
var menu = new ChatMenu("My wonderful menu title");

menu.AddMenuOption("Live", OnSelectHandler);
menu.AddMenuOption("Laugh", OnSelectHandler);
menu.AddMenuOption("Love", OnSelectHandler);

menu.AddMenuOption("Exit", OnSelectExit);

ChatMenus.OpenMenu(player, menu);
```

After:
```c#
var menu = new ChatMenu("My wonderful menu title");

menu.AddMenuOption("Live", OnSelectHandler);
menu.AddMenuOption("Laugh", OnSelectHandler);
menu.AddMenuOption("Love", OnSelectHandler);

menu.AddMenuOption("Exit", OnSelectExit);

MenuManager.OpenChatMenu(player, menu);
```

```c#
// This is because to keep the menu showing to a player we need to add an OnTick listener to do it.
// plugin = this; when in the plugin class.
var plugin = this;

var menu = new CenterHtmlMenu("My wonderful menu title");

menu.AddMenuOption("Live", OnSelectHandler);
menu.AddMenuOption("Laugh", OnSelectHandler);
menu.AddMenuOption("Love", OnSelectHandler);

menu.AddMenuOption("Exit", OnSelectExit);

MenuManager.OpenCenterHtmlMenu(plugin, player, menu);
```

```c#
var menu = new ConsoleMenu("My wonderful menu title");

menu.AddMenuOption("Live", OnSelectHandler);
menu.AddMenuOption("Laugh", OnSelectHandler);
menu.AddMenuOption("Love", OnSelectHandler);

menu.AddMenuOption("Exit", OnSelectExit);

MenuManager.OpenConsoleMenu(player, menu);
```

I've also maintained backwards compatibility so this still works (but will throw warnings @ compile time):
```c#
var menu = new ChatMenu("My wonderful menu title");

menu.AddMenuOption("Live", OnSelectHandler);
menu.AddMenuOption("Laugh", OnSelectHandler);
menu.AddMenuOption("Love", OnSelectHandler);

menu.AddMenuOption("Exit", OnSelectExit);

ChatMenus.OpenMenu(player, menu);
```


managed/CounterStrikeSharp.API/Modules/Menu/CenterHtmlMenu.cs
```diff
@@ -0,0 +1,102 @@
...
+            MenuOptions.Add(option);
+            return option;
+        }
+    }
+
+    public class CenterHtmlMenuInstance: BaseMenu
+    {
+        public CenterHtmlMenuInstance(BasePlugin plugin, CCSPlayerController player, IMenu menu) : base(player, menu)
+        {
+            plugin.RegisterListener<Core.Listeners.OnTick>(Display);
```
yonilerner, line 42: I guess `OnTick` is necessary?

B3none: It's how the previous guy implemented it CenterHtml Menus, this may also be required for `CenterMenu` too

B3none:
I think the build error is failing over a non-issue here:
```
Error: /home/runner/.nuget/packages/microsoft.dotnet.apicompat.task/7.0.404/build/Microsoft.DotNet.ApiCompat.ValidateAssemblies.Common.targets(16,5): error CP0002: Member 'CounterStrikeSharp.API.Modules.Menu.ChatMenuOption.ChatMenuOption(string, bool, System.Action<CounterStrikeSharp.API.Core.CCSPlayerController, CounterStrikeSharp.API.Modules.Menu.ChatMenuOption>)' exists on ./ApiCompat/v133.dll but not on obj/Debug/net7.0/CounterStrikeSharp.API.dll [/home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/CounterStrikeSharp.API.csproj]
Error: /home/runner/.nuget/packages/microsoft.dotnet.apicompat.task/7.0.404/build/Microsoft.DotNet.ApiCompat.ValidateAssemblies.Common.targets(16,5): error CP0002: Member 'CounterStrikeSharp.API.Modules.Menu.ChatMenu.AddMenuOption(string, System.Action<CounterStrikeSharp.API.Core.CCSPlayerController, CounterStrikeSharp.API.Modules.Menu.ChatMenuOption>, bool)' exists on ./ApiCompat/v133.dll but not on obj/Debug/net7.0/CounterStrikeSharp.API.dll [/home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/CounterStrikeSharp.API.csproj]
Error: /home/runner/.nuget/packages/microsoft.dotnet.apicompat.task/7.0.404/build/Microsoft.DotNet.ApiCompat.ValidateAssemblies.Common.targets(16,5): error CP0002: Member 'CounterStrikeSharp.API.Modules.Menu.ChatMenuInstance.ChatMenuInstance(CounterStrikeSharp.API.Core.CCSPlayerController, CounterStrikeSharp.API.Modules.Menu.ChatMenu)' exists on ./ApiCompat/v133.dll but not on obj/Debug/net7.0/CounterStrikeSharp.API.dll [/home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/CounterStrikeSharp.API.csproj]
Error: /home/runner/.nuget/packages/microsoft.dotnet.apicompat.task/7.0.404/build/Microsoft.DotNet.ApiCompat.ValidateAssemblies.Common.targets(16,5): error CP0002: Member 'CounterStrikeSharp.API.Modules.Menu.ChatMenus.OpenMenu(CounterStrikeSharp.API.Core.CCSPlayerController, CounterStrikeSharp.API.Modules.Menu.ChatMenu)' exists on ./ApiCompat/v133.dll but not on obj/Debug/net7.0/CounterStrikeSharp.API.dll [/home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/CounterStrikeSharp.API.csproj]
Error: Process completed with exit code 1.
```

I suspect it's failing because I've changed some of the types of the ChatMenu stuff to interface types (which are the same structure since they're based off of the ChatMenu structure)

Cruze03:
!7 for back would be nice instead of !1

B3none:
> !7 for back would be nice instead of !1

I'm trying to make it the same as what is already in the existing `ChatMenu` implementation.

ItsChase88:
Would it be possible, for all the implementations besides ChatMenu, to track player buttons so they just push 1, 2, 3, etc rather than forcing players to type !x?

For the chat menu, it makes sense. But because we can display the menu elsewhere on the HUD we'll have a normal menu again if we can track players' buttons. This is assuming dynamic hooks have been added I'm not too sure if this has been finished or not yet.

Dliix66 (approved):

CHR15cs:
Is the Html menu staying open after a single selection or does it just close like the chat menu?
I think an optional parameter to allow multiple selections would be nice.

B3none:
> Is the Html menu staying open after a single selection or does it just close like the chat menu? I think an optional parameter to allow multiple selections would be nice.

The menu will close after a selection is made. Once this is merged you can make a feature request for it 👍

Dliix66:
Agreed! I will if I have some time ;)

B3none:
Yeah I need to set a limit for the title of a `CenterHtmlMenu`:
image

managed/CounterStrikeSharp.API/Modules/Menu/CenterHtmlMenu.cs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    public class CenterHtmlMenuInstance : BaseMenuInstance
+    {
+        private readonly BasePlugin _plugin;
+
+        public CenterHtmlMenuInstance(BasePlugin plugin, CCSPlayerController player, IMenu menu) : base(player, menu)
+        {
+            _plugin = plugin;
+            RemoveOnTickListener();
```
zonical, line 65: Would this not cause issues with plugins that already have `OnTick` defined?

managed/CounterStrikeSharp.API/Modules/Menu/IMenu.cs
```diff
@@ -0,0 +1,44 @@
...
+        protected int Page { get; }
+        protected int CurrentOffset { get; }
+        protected int NumPerPage { get; }
+        protected Stack<int> PrevPageOffsets { get; }
+        
+        public void NextPage();
+        public void PrevPage();
+        public void Reset();
+        public void Display();
+        public void OnKeyPress(CCSPlayerController player, int key);
```
zonical, line 42: While this technically isn't _wrong_, I would recommend renaming this function to something else (maybe `OnOptionSelected`) to avoid confusion between this implementation of menus and the old SourceMod style of menus. There currently isn't a way to recreate those menus or to detect key presses. Maybe when we get support for UserMessages in the future 🤞

managed/CounterStrikeSharp.API/Modules/Menu/MenuManager.cs
```diff
@@ -0,0 +1,76 @@
...
+    
+    public static void OpenConsoleMenu(CCSPlayerController player, ConsoleMenu menu)
+    {
+        ResetMenus(player);
+        
+        ActiveMenus[player.Handle] = new ConsoleMenuInstance(player, menu);
+        ActiveMenus[player.Handle].Display();
+    }
+
+    public static void OnKeyPress(CCSPlayerController player, int key)
```
zonical, line 69: See related comment for IMenu.

managed/CounterStrikeSharp.API/Modules/Menu/CenterHtmlMenu.cs
Dliix66: Yes good catch, as it calls `_plugin.RemoveListener("OnTick", onTick);` the whole plugin would stop receiving `OnTick`

zonical:
Most of the new implementation looks good to me, well done.

The only concern I have is the removal of the `OnTick` with `RemoveOnTickListener` as some plugins will depend on its functionality. [Looking at RegisterListener<T> in BasePlugin.cs](https://github.com/roflmuffin/CounterStrikeSharp/blob/765c56a38af0f0e5a7a8f4fcb9786cb56461c8bc/managed/CounterStrikeSharp.API/Core/BasePlugin.cs#L339), it doesn't seem possible to register multiple listeners at once as `listenerName` is set by T's type. This might require a breaking change in the Core API ( cc: @roflmuffin ) where developers can define multiple listeners in a plugin and assign them to a name. A function would also have to be implemented to remove listeners at runtime as well.

For example:
```c#
RegisterListener<Listeners.OnTick>(() => { ... }, "HTMLHandler");
RemoveListener<Listeners.OnTick>("HTMLHandler");
```

managed/CounterStrikeSharp.API/Modules/Menu/CenterHtmlMenu.cs
B3none: I don’t think it clashes, the listener has a key and this function just removes the display function from the array of OnTick listeners afaik. I can test it to make sure the other OnTick listeners aren’t cleared but I’m pretty sure they won’t be.

managed/CounterStrikeSharp.API/Modules/Menu/IMenu.cs
B3none: It’s still called OnKeyPress to maintain backwards compatibility, plus once we figure out how to detect the number keys it’ll make sense and won’t need refactoring.

Dliix66:
> Most of the new implementation looks good to me, well done.
> 
> The only concern I have is the removal of the `OnTick` with `RemoveOnTickListener` as some plugins will depend on its functionality. [Looking at RegisterListener in BasePlugin.cs](https://github.com/roflmuffin/CounterStrikeSharp/blob/765c56a38af0f0e5a7a8f4fcb9786cb56461c8bc/managed/CounterStrikeSharp.API/Core/BasePlugin.cs#L339), it doesn't seem possible to register multiple listeners at once as `listenerName` is set by T's type. This might require a breaking change in the Core API ( cc: @roflmuffin ) where developers can define multiple listeners in a plugin and assign them to a name. A function would also have to be implemented to remove listeners at runtime as well.
> 
> For example:
> 
> ```cs
> RegisterListener<Listeners.OnTick>(() => { ... }, "HTMLHandler");
> RemoveListener<Listeners.OnTick>("HTMLHandler");
> ```

I did implement such a thing on my WCS plugin, instead of using a key, it uses an object, so you can pass the object that subscribed to it. It works fine, although if implemented in the core, could be improved as I used what was publicly available.
We could discuss it on Discord before working on a PR?

Dliix66:
Also, even if we skip the unregister, that means plugins will not be able to subscribe to OnTick as only one of the handler will actually be triggered.

B3none:
I tested your concerns about the on tick handler; find a video of it working / not interfering here:

https://github.com/roflmuffin/CounterStrikeSharp/assets/24966460/5aa98807-b629-4c93-8c2e-c067bb9e0ebf

zonical:
Sweet, no concerns from me then. LGTM!

zwolof (approved):

managed/CounterStrikeSharp.API/Modules/Menu/CenterHtmlMenu.cs
roflmuffin: Yeah it should be using the actual delegate passed through to remove; so it can only remove listeners it has the actual underlying method being called, so this shouldn't cause any issues

PR #269, opened 2024-01-18
Fix css_plugins commands number of args check

abnerfs:
Checks for number of args in css_plugins commands (css_plugins reload / start / unload), they are checking for < 2 args, they should check for < 3 args since the css_plugins command count as an argument.

Before my change:
image

After my change:
image


B3none:
Old behaviour:
image

New behaviour:
image

B3none (approved):

PR #278, opened 2024-01-20
Admin manager improvements

zonical:
A few additions after a conversation on Discord:
- Updated comments for AddPlayerToGroup 
- Added player controller equivalent for GetPlayerAdminData and RemovePlayerAdminData
- Added GetPlayerImmunity
- Prevent HLTV player controller from being passed into admin-related functions
- Changed TargetType.GroupAll TargetPredicate to exclude HLTV

I've decided to exclude the HLTV player from these functions as they do not have a physical presence in the world and shouldn't be manipulated with commands.


B3none (dismissed):
Seems like it maintains backwards compatibility. LGTM 👍

PR #284, opened 2024-01-22
HTML Menu improvements

Dliix66:
Here are the changes:
- Removed 1 max option per page on HTML display to avoid truncated option (only the "!9 -> Close" could be truncated)
- Moved "Prev Page" from "!1" to "!7" to match old Radio Menus' behavior for all menus
- Added grey color for disabled options in HTML
- Made "Prev Page" option yellow just like next page

And here is the result:

1) Max options in a menu
image

2) Prev Option
image

3) Disabled options
image


B3none (approved):
These changes look legit, thanks! 🙏

roflmuffin (approved):

PR #287, opened 2024-01-23
Tidy CCSPlayerController

B3none:
- Removed redundant calls to "this."


KillStr3aK:
the project indeed doesn't follow the 'this' usage in the codestyle, however its only considered as redundant in given contexts

roflmuffin (approved):
Will probably add an editor config for this eventually

PR #290, opened 2024-01-26
Made player.Teleport() teleport the PlayerPawn as it was confusing for plugin developers

B3none:
I personally lost 30 mins debugging this only to find that I wanted to teleport the PlayerPawn instead 😢


KillStr3aK:
could you elaborate? what have you tried and what happened?

teleporting the playerpawn doesnt seems totally good to me (assume that nothing has changed since the last time I've been on it)

B3none:
> could you elaborate? what have you tried and what happened?
> 
> teleporting the playerpawn doesnt seems totally good to me (assume that nothing has changed since the last time I've been on it)

If people want to teleport the player around calling `player.Teleport()` won't do anything. You must call the teleport function on the PlayerPawn entity for the player to be moved

roflmuffin:
I know we have added this for convenience for other methods (like `GiveNamedItem`), but I think this is actually a design mistake on my part and I would like to deprecate these methods (and point to using the playerpawn methods directly), so would like to prevent creation of new ones at this stage.

PR #294, opened 2024-01-28
Improved menu closing

B3none:
Allow plugins to close the active menu in the MenuManager, also updated OnKeyPress to utilise the GetActiveMenu function and added the .Close() function to a base menu instance

### DIFF
--- a/managed/CounterStrikeSharp.API/Modules/Menu/BaseMenu.cs
+++ b/managed/CounterStrikeSharp.API/Modules/Menu/BaseMenu.cs
@@ -121,7 +121,7 @@ namespace CounterStrikeSharp.API.Modules.Menu
             PrevPageOffsets.Clear();
         }
         
-        public virtual void Close()
+        public void Close()
         {
             MenuManager.CloseActiveMenu(Player);
         }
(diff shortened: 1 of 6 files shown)


partiusfabaa:
It might be easier to add a parameter, like `var menu = new ChatMenu("title", true or false);`, responsible for determining whether the menu will close after pressing the button. Alternatively, in the handler, you can `return true or false`. Depending on the returned value, the menu will be closed or remain open. I believe these two options are ideal.

roflmuffin:
To fix the API compatibility issues you would need to: define the 3rd parameter not as optional, but as an overloaded method (since optional params are a compile time thing), i.e. instead of `public static void OpenChatMenu(CCSPlayerController player, ChatMenu menu, bool closeOnSelect = true)` you leave the old one`public static void OpenChatMenu(CCSPlayerController player, ChatMenu menu), which just calls the new one `public static void OpenChatMenu(CCSPlayerController player, ChatMenu menu, bool closeOnSelect)` with `true`.

You also need to provide a default implementation for `Close()` that does nothing (at the interface level) which keeps the API compatibility in check I believe.

B3none:
Converted this back to a draft because I'm not happy with implementing the `CloseOnSelect` in the constructor and I believe it would be better as a menu setter. `myMenu.SetCloseOnSelect(true);` defaulting the behaviour to true (as it was previously implemented).

B3none:
I opted to use a PostSelectAction enum instead: 
```c#
public enum PostSelectAction
{
    Close,
    Reset,
    Nothing
}
```

How it looks to people using a menu in a plugin:
```c#
menu.PostSelectAction = PostSelectAction.Close;
```

roflmuffin (approved):

PR #297, opened 2024-01-29
Update .gitmodules to use https references rather than git reference

B3none:


B3none:
Is there any reason why this has not been merged?

roflmuffin (approved):

PR #311, opened 2024-02-07
Update gamedata for Call To Arms Update

zonical:
**I cannot stress this enough. If you are a server owner, please wait until an official release is pushed.**
Updated signatures from CS2Fixes as well as various sources on Discord. `CCSPlayerController_Respawn` is causing invalid function pointer errors, but everything else looks okay from my initial testing on Linux. **Not tested on Windows.**


Cruze03:
Untested but as Poggu said somewhere, we might need to change [this](https://github.com/roflmuffin/CounterStrikeSharp/compare/main...Cruze03:CounterStrikeSharp:main)

zonical:
This introduces a breaking change to CCSPlayerController.Respawn which swaps out `CCSPlayerPawn_Respawn` for `CBasePlayerController_SetPawn`, which should guarantee that player pawns will be properly respawned. Plugin developers should update to the latest version when it is released and stop using `CCSPlayerPawn.Respawn`

managed/CounterStrikeSharp.API/Core/Model/CCSPlayerPawn.cs
```diff
@@ -5,15 +5,17 @@ namespace CounterStrikeSharp.API.Core;
...
\ No newline at end of file
+	[Obsolete]
+	public void Respawn()
+	{
+		if (!Controller.IsValid) return;
+		if (!Controller.Value.IsValid) return;
+
+		// bit of a bodge ¯\_(ツ)_/¯
+		Controller.Value.As<CCSPlayerController>().Respawn();
+	}
```
roflmuffin, line 20:
```suggestion
	[Obsolete("Use CCSPlayerController.Respawn() instead")]
	public void Respawn()
	{
		if (!Controller.IsValid) return;
		if (!Controller.Value.IsValid) return;

		Application.Instance.Logger.LogWarning("CCSPawn.Respawn is deprecated and does nothing, use CCSPlayerController.Respawn instead");
	}
```

roflmuffin (approved):

roflmuffin (approved):

PR #333, opened 2024-02-21
feat: Include description in Discord release notification

switz:
I have not tested this as it involves a release and discord token, but followed the requisite docs. This should include the release description in the discord notification so we get more clarity into what changed in the release.


switz:
nevermind, the release actions actually do not include the release description. I don't see how to pull this without fetching the release itself in a new step. Will close this for now.

roflmuffin:
Yeah I've been trying to work out a nice way to do this as well but it seemed more effort than it's worth

PR #342, opened 2024-02-25
Draft: Feature/shipped plugins

B3none:
Plugins left to convert:
- [ ] [admin-flatfile.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/admin-flatfile.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [admin-sql-prefetch.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/admin-sql-prefetch.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [admin-sql-threaded.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/admin-sql-threaded.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [adminhelp.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/adminhelp.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [adminmenu.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/adminmenu.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [antiflood.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/antiflood.sp)
    - [x] Port plugin @B3none
    - [ ] Documentation

- [ ] [basebans.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/basebans.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [basechat.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/basechat.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [basecomm.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/basecomm.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [basecommands.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/basecommands.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [basetriggers.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/basetriggers.sp)
    - [ ] Port plugin
    - [ ] Documentation

- [ ] [basevotes.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/basevotes.sp)
    - [ ] Port plugin
    - [ ] Documentation

[...]

PR #344, opened 2024-02-25
Add RockTheVote to shipped plugins

abnerfs:
Add https://github.com/abnerfs/cs2-rockthevote as a submodule in shipped plugins


B3none:
This is not what we're after unfortunately. There are talks of potentially creating a CounterStrikeSharp organisation to we may be able to do it there at a later date, but for now I'd like the plugin to be in the version control of this repo.

zonical:
Marking as closed until the discussion in #342 is resolved. Depending on the approach that is taken, this PR might be reopened.

Issue #327, opened 2024-02-16
Make player-grabbing functions safer


zonical:
I wanted to do this myself before I left for university, but I'll offer this task to someone else.

A lot of issues when developing plugins that manipulate players come down to not validating player controller and player pawn entities properly. We should be building these validity checks into the Utilities functions instead of writing wrapper libraries, which I have done in this case.

The minimum validity checks should be:
- The player controller entity is valid
- UserId and Slot
- The player is fully connected to the server
- The player is not HLTV (perhaps add a new function to get the HLTV player)
- The active pawn (spectator or player) is valid

roflmuffin:
I'm wondering if we can re-purpose the `TargetType` from the target module for this, where the user can supply their own simplistic filter:

```csharp
switch (targetType)
        {
            case TargetType.TeamCt:
                return TeamNum == (byte)CsTeam.CounterTerrorist;
            case TargetType.TeamT:
                return TeamNum == (byte)CsTeam.Terrorist;
            case TargetType.TeamSpec:
                return TeamNum == (byte)CsTeam.Spectator;
            case TargetType.GroupAll:
                return !IsHLTV;
            case TargetType.GroupBots:
                return IsBot;
            case TargetType.GroupHumans:
                return !IsBot && !IsHLTV;
            case TargetType.GroupAlive:
                return PlayerPawn is { IsValid: true, Value.LifeState: (byte)LifeState_t.LIFE_ALIVE };
            case TargetType.GroupDead:
                return PlayerPawn is { IsValid: true, Value.LifeState: (byte)LifeState_t.LIFE_DEAD or (byte)LifeState_t.LIFE_DYING };
            default:
                return false;
        }
```

biggestmannest:
one option would be interfaces. then you could make extensions for them :)

```cs
interface IEntityValid
{
  public bool IsValid;
}

public static class EntityExtensions
{
  public static bool TryGetEntity<T>(this T? @this, [NotNullWhen(true)] out T? entity) where T : NativeEntity, IEntityValid
  {
    entity = null;

    if (@this is null || !@this.IsValid)
    {
      return false;
    }

    entity = @this;

    return true;
  }
}
```

PR #349, opened 2024-02-27
feat: Improve player validation

zonical:
Addresses #327. Similar in nature to Poggie's entity validation PR, this adds validation checks for grabbing players.


roflmuffin:
Sorry, I'm not sure if I fundamentally agree with the changes in this PR; we don't gain a huge amount of value just shifting the player methods from one static class to another, aside from annoying people with an API change. Eventually I could see us adding these as factories from the `CCSPlayerController` class as `CCSPlayerController.FromIndex()` etc.

The guard statement is not really a guard statement, as these are conventionally a statement that either does nothing (passes), or throws an exception, rather than returning a boolean. I think the name `Ready` is a bit ambiguous too.

I'm also unsure if its great idea to throw an exception from the `GetPlayerAuthorizedSteamID` native, I feel like this probably would cause a lot of load time errors on our side as we probably just return no permissions for this user until their steam authenticated for instance.

PR #365, opened 2024-03-06
feat: use concurrent queue for next frame & world update tasks

### LINKED ISSUE #354: NextFrame/NextWorldUpdate fails when nested
It is not currently possible to nest NextFrame or NextWorldUpdate. 

```csharp
Server.NextFrame(() =>
{
  Server.NextFrame(() =>
  {
    Logger.LogInformation("Crash!");
  });
});
```

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at CounterStrikeSharp.API.Core.Helpers.InvokeNative(IntPtr ptr)
   at CounterStrikeSharp.API.Core.ScriptContext.InvokeNativeInternal() in C:\Users\Michael\source\repos\CounterStrikeSharp\managed\CounterStrikeSharp.API\Core\ScriptContext.cs:line 114
   at CounterStrikeSharp.API.Core.ScriptContext.Invoke() in C:\Users\Michael\source\repos\CounterStrikeSharp\managed\CounterStrikeSharp.API\Core\ScriptContext.cs:line 103
   at CounterStrikeSharp.API.Core.NativeAPI.QueueTaskForNextWorldUpdate(InputArgument callback) in C:\Users\Michael\source\repos\CounterStrikeSharp\managed\CounterStrikeSharp.API\Core\API.cs:line 520
   ```

### PR DESCRIPTION
**Pull request by roflmuffin**
- Uses MPMC queue for frame related tasks, which allows nested next frame calls and processes up to 1024 tasks per frame.
- Adds Task completion to next frame helpers, which allows them to be used in an async context and awaited which provides more reasonable behaviour.
- Fixes thread safety of `FunctionReference` identifiers, as these were not behind a lock, so multiple threads creating new function references (i.e. `Server.NextFrame`) would cause id dictionary collisions and garbage collector would free the underlying callback, causing a crash.
- Forces a new callback for all `FunctionReference` created with a single use lifetime, previously if your lambda was re-used (and did not capture variables), it would be looked up in a dictionary and the same function callback would be used, causing misleading behaviour. Single use lambdas now always point to a fresh callback pointer.
- Significantly refactors FunctionReference to better outline its purpose

Closes #354

### DIFF
--- a/libraries/moodycamel/concurrentqueue.h
+++ b/libraries/moodycamel/concurrentqueue.h
@@ -1,0 +1,3747 @@
+// Provides a C++11 implementation of a multi-producer, multi-consumer lock-free queue.
+// An overview, including benchmark results, is provided here:
+//     http://moodycamel.com/blog/2014/a-fast-general-purpose-lock-free-queue-for-c++
+// The full design is also described in excruciating detail at:
+//    http://moodycamel.com/blog/2014/detailed-design-of-a-lock-free-queue
+
+// Simplified BSD license:
+// Copyright (c) 2013-2020, Cameron Desrochers.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+//
+// - Redistributions of source code must retain the above copyright notice, this list of
+// conditions and the following disclaimer.
+// - Redistributions in binary form must reproduce the above copyright notice, this list of
+// conditions and the following disclaimer in the documentation and/or other materials
+// provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by roflmuffin** [comment]
Tested this with pretty jank code, and confirmed the output is as expected, 1024 calls per frame until the queue is emptied.

```csharp
// Queue next world updates from as many threads as possible
var times = new Dictionary<float, int>();
var tasks = new List<Task>();
for (int i = 0; i < 10000; i++)
{
    // Create a new task that will run on a random thread
    tasks.Add(Task.Run(async () =>
    {
        await Server.NextWorldUpdate(() =>
        {
            // Track the server (main thread) time when the task is executed
            times[Server.CurrentTime] = times.TryGetValue(Server.CurrentTime, out var time) ? time + 1 : 1;
        });
    }));
}

await Task.WhenAll(tasks);


XunitTestPlugin.Instance.Logger.LogInformation("{Times}", times);
```

`{"14.25": 1024, "14.28125": 1024, "14.3125": 1024, "14.328125": 1024, "14.34375": 1024, "14.359375": 1024, "14.375": 1024, "14.40625": 1024, "14.421875": 1024, "14.4375": 784}`

PR #373, opened 2024-03-10
minor menu changes

partiusfabaa:
### 1. Added option to disable the exit button

```cs
var menu = new ChatMenu("TEST");
menu.ExitButton = true;

menu.AddMenuOption("Test", (controller, option) => Server.PrintToChatAll(option.Text));

MenuManager.OpenChatMenu(player, menu);
```

### 3. Added exit button to console and chat menu


roflmuffin (changes requested):
Could we change the `ExitButton` to default to `false` for ChatMenu? This matches previous 
behaviour.

It looks like in my testing, on a HTML menu once you go to the next page, the `!9 close` option disappears because there is not enough room, though you can still do !9 to close.

roflmuffin (approved):

K4mYY:
Can you also change the buttons and align them with csgo/css please ?
!8 previous, !9 next, !0 exit

PR #374, opened 2024-03-11
feat: add `Server.RunOnTick` method which allows tick scheduling

### LINKED ISSUE #367: Add RunInNTicks or RunOnTick
It would be great to be able to do things like "run in 3 ticks."

```csharp
RunInNTicks(3, () => Logger.LogInformation("wooo, 3 ticks!"));
```

and/or

```csharp
var futureTick = Server.TickedTime + 3;
RunOnTick(futureTick, () => Logger.LogInformation("wooo, 3 ticks!"));
```

This has been discussed briefly here: https://github.com/roflmuffin/CounterStrikeSharp/issues/354#issuecomment-1980129038

roflmuffin:
Adds `RunOnTick` and `RunOnTickAsync` methods which allow you to queue a task for a specific game frame. Closes #367 

Example usage:

```csharp
Server.RunOnTick(Server.TickCount + 64, () => Console.WriteLine("Hello World"));
```

The async overload simply returns a `Task` which can be awaited which resolves once the callback has executed.

### DIFF
--- a/managed/CounterStrikeSharp.API/Server.cs
+++ b/managed/CounterStrikeSharp.API/Server.cs
@@ -28,12 +28,18 @@ namespace CounterStrikeSharp.API
 {
     public class Server
     {
+        /// <summary>
+        /// Duration of a single game tick in seconds, based on a 64 tick server (hard coded in CS2).
+        /// </summary>
         public static float TickInterval => 0.015625f;
 
+        /// <summary>
+        /// Executes a command on the server, as if it was entered from the console.
+        /// </summary>
+        /// <param name="command"></param>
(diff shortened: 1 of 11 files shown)


managed/CounterStrikeSharp.API/Server.cs
```diff
@@ -28,19 +28,51 @@ namespace CounterStrikeSharp.API
 {
     public class Server
     {
-        public static float TickInterval => NativeAPI.GetTickInterval();
+        public static float TickInterval => 0.015625f;
```
roflmuffin, line 31: This is hardcoded on native side as well, just saving a native call here for no real reason.

PR #379, opened 2024-03-14
Add entities pre and post think

Yarukon:
This will called every frame before/after entities think (maybe it will give some benefit to some features?)

Example:
```C#
RegisterListener<Listeners.OnServerPreEntityThink>(() =>
{
    // do something
});

RegisterListener<Listeners.OnServerPostEntityThink>(() =>
{
    // do something
});
```


roflmuffin:
I'm not against this addition, but curious what problems people would use using this over regular tick and pre world update callbacks?

Yarukon:
> I'm not against this addition, but curious what problems people would use using this over regular tick and pre world update callbacks?

maybe update player state before entity think?

roflmuffin:
I think I would prefer to implement this as a per-entity listener instead, and subscribed to similarly to entity io output, as I think that would probably be more useful

PR #386, opened 2024-03-20
Add EmitSound

Yarukon:
Allows to play sound on entity.


Yarukon:
linux build dommed, turn to draft

Yarukon:
closing for now

B3none:
Awesome! This will be a nice addition to the project 👍

chenbuyi2019:
test good on Windows:
```c#
ent.EmitSound("jungle_male.afm1_thankful_01");
```
it has to be a sound event name, not a vsnd filename.
and which means SoundLevel, volume etc params are not working, since they are defined in sound events already.
Filter is working good too.

roflmuffin:
Do you think we should just omit the volume, pitch etc. if they are not actually working at the moment anyway? And this can be added in a later API once we have a way to modify the volume of an existing sound, similar to this implementation: https://gist.github.com/Metapyziks/88745748d90f8955e60b9831a7214cd2

chenbuyi2019:
> Do you think we should just omit the volume, pitch etc. if they are not actually working at the moment anyway? And this can be added in a later API once we have a way to modify the volume of an existing sound, similar to this implementation: https://gist.github.com/Metapyziks/88745748d90f8955e60b9831a7214cd2

You're right, we should have a better way to emit sound with real volume changing. 
Sometimes, re-editing sound event file is really a mess.

Metapyziks:
I've stalled a bit on being able to change sound parameters after emitting, I can't get it to work with custom sound events using MultiAddonManager.

Dliix66:
For now the only way to manage volume is to use the Hammer config when creating the map with the sounds. I remember somebody posted some code to update the volume but it was pretty unstable IIRC.

I'd say this is a cool feature to have and we can update it once we found a way for the volume ;)

Yarukon:
draft this fn

Dliix66:
> draft this fn

Is this working on the latest update? I’ll give it a try tonight

Yarukon:
> > draft this fn
> 
> Is this working on the latest update? I’ll give it a try tonight

works on windows, you can try on linux if you have linux enviroment

Yarukon:
closing

PR #388, opened 2024-03-21
Fix backwards compatibility in DynamicHook

ipsvn:
Re-add the unused parameter with a overload because it breaks plugins not updated to also remove said parameter

### DIFF
--- a/managed/CounterStrikeSharp.API/Modules/Memory/DynamicFunctions/DynamicHook.cs
+++ b/managed/CounterStrikeSharp.API/Modules/Memory/DynamicFunctions/DynamicHook.cs
@@ -14,6 +14,7 @@ public class DynamicHook : NativeObject
         return NativeAPI.DynamicHookGetParam<T>(Handle, (int)typeof(T).ToValidDataType(), index);
     }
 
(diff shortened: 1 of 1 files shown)


B3none (approved):
Looks good, we like to try and ensure backwards compatibility with every change so I guess this one slipped the net!

partiusfabaa:
Then why was it necessary to accept this PR #334? So many plugins where GetReturn with an index is used?

B3none:
> Then why was it necessary to accept this PR(#334)? So many plugins where GetReturn with an index is used?

It’s about maintaining backwards compatibility, as far as I’m aware your PR shouldn’t have been merged without adding this change to maintain backwards compatibility

ipsvn:
> Then why was it necessary to accept this PR(#334)? So many plugins where GetReturn with an index is used?

I don't believe that many, so it isn't a massive issue, but it would've been nice if this was included in the original PR to give some time before it was removed completely and just breaking any existing plugins using it. Anyone using it has probably fixed it by now though, so it's probably not that important. ¯\_(ツ)_/¯

PR #399, opened 2024-04-06
Add teleports methods

partiusfabaa:
1. Added `Zero` field to `Vector`, `QAngle` and `Angle`. To prevent unnecessary allocations. 
2. Added additional `Teleport` methods in case the user does not want to change angles or velocity.


roflmuffin:
Thoughts on instead creating a method `public void Teleport(Vector? position = null, QAngle? angles = null, Vector? velocity = null)` and then defaulting based on their null value, then the consumer just uses named args like `.Teleport(position: x);`

We should initialise the value of `public static readonly Vector Zero` in `Application.cs` or `Bootstrap.cs` so that we do this safely at app startup (when we know natives are allowed to be called) rather than at usage time.

partiusfabaa:
this `Zero` field will only be used by plugins and it will not be initialized until something accesses the class. I don't think it's worth it

partiusfabaa:
> We should initialise the value of `public static readonly Vector Zero` in `Application.cs` or `Bootstrap.cs` so that we do this safely at app startup (when we know natives are allowed to be called) rather than at usage time.

this `Zero` field will only be used by plugins and it will not be initialized until something accesses the class. I don't think it's worth it

roflmuffin (approved):

PR #407, opened 2024-04-10
Make signature support IDA/x64Dbg style

Nukoooo:
See [discussion](https://github.com/roflmuffin/CounterStrikeSharp/discussions/404) for more info.

TLDR: Make developers' life easier so they don't have to convert their signatures(``48 8B ?``) to code style(``\x48\x8B\x2A``) every time when they make a signature

### What's changed
- Signature now supports IDA/x64Dbg style (``48 8B ? ? ? ? ?``)
- Removed duplicated code
- Switched to C++20 (because of ``std::views::split``)
- ``FindSignature`` now uses ``std::find`` and ``std::equal`` for better readability


roflmuffin:
LGTM, what platforms have you tested this on?

Nukoooo:
I only tested on Windows. but I also tried to compile with GCC 13.2.1 and Clang 17.0.6 on my Arch WSL, while GCC ccompiled successfully without any error, clang required to remove ``register`` keywords in ``hl2sdk-cs2/tier1/generichash.cpp`` to make it work

Nukoooo:
Just took a look at why the build failed on linux, it turned out ``std::string_view(split.begin(), split.end())`` only exists in GCC 12.1 and above and Clang 16.0.0 and above see [here](https://gcc.godbolt.org/z/T7e8v8WjK). So it might need to upgrade the compiler that workflow uses newer compiler.

roflmuffin:
Tested on Linux using latest build artifact ✅

roflmuffin (approved):

PR #414, opened 2024-04-12
Copy bytes of a module to prevent signature scanning failures and some minor changes

Nukoooo:
### Why
With the current implementation of ``FindSignature``, we scan bytes in memory, which can get patched or modified during runtime, and if we try to scan bytes that are patched then it results in a failure. Here is the minimum code to reproduce:
```cs
private MemoryFunctionWithReturn<CPlayer_ObserverServices, IntPtr, bool> CPlayerObserverService_OnClientCommand;

public override void OnAllPluginsLoaded(bool hotReload)
{
    var signature = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
        ? @"\x48\x89\x5C\x24\x2A\x48\x89\x6C\x24\x2A\x48\x89\x74\x24\x2A\x57\x48\x83\xEC\x2A\x83\xBA\x2A\x2A\x2A\x2A\x2A\x48\x8D\x2D\x2A\x2A\x2A\x2A\x48\x8B\xF2\x48\x8B\xF9"
        : @"\x55\x48\x89\xE5\x41\x57\x41\x56\x41\x55\x4C\x8D\x2D\x2A\x2A\x2A\x2A\x41\x54\x49\x89\xFC\x53\x48\x89\xF3";

    var address = NativeAPI.FindSignature(Addresses.ServerPath, signature);

    // Returns the address
    Logger.LogInformation($"[Prehook] CPlayerObserverService_OnClientCommand: 0x{address:X}");

    CPlayerObserverService_OnClientCommand = new(signature, Addresses.ServerPath);
    CPlayerObserverService_OnClientCommand.Hook(Handler, HookMode.Pre);

    address = NativeAPI.FindSignature(Addresses.ServerPath, signature);

    // If we find the signature after it is hooked, it returns 0.
    // Because the first 5 bytes (assuming it is within 2GB range) are replaced with "0xE8 0xDE 0xAD 0xBE 0xEF"
    // and the bytes we are looking for, take Linux signature as an example, are "0x55 0x48 0x89 0xE5 0x41"
    // hence FindSignature fails to find the bytes we need
    Logger.LogInformation($"[PostHook] CPlayerObserverService_OnClientCommand: 0x{address:X}");
}

private HookResult Handler(DynamicHook arg)
{
    return HookResult.Continue;
}
```

### What's changed in this PR

- Cache game libraries before loading the core
    - Copy bytes from executable sections and FindSignature will now search through saved bytes
    - For ``engine2`` and ``server`` libraries, they read bytes from disk instead of memory to prevent some functions gets hooked before loading the plugin framework
    - It won't cache libraries without ``CreateInterface`` symbol
- Cache symbols/exports for better performance (can hash symbol name for better performance)
- Cache interfaces for better performance (can also hash interface name for better performance)
- Fix error when using macro ``CSSHARP_CORE_DEBUG``
- Add and fix a log message when it fails to convert hex string to bytes


NaiJii:
This is good, I did something similar (but didn't cache from files) and it improves sigscans speed quite a bit.

roflmuffin:
I'm assuming this means then if you call a virtual function that has been hooked, you will find the newly trampolined address and that works as expected?

Nukoooo:
> the newly trampolined address

I'm not sure what this means, if it means the address of a hooked function then yes, this PR addresses this issue.

src/core/globals.cpp
```diff
@@ -28,6 +28,7 @@
 namespace counterstrikesharp {
 
 namespace modules {
+std::unordered_map<std::string, std::unique_ptr<CModule>> moduleList {};
```
laper32, line 31: it's a map but it named `module**List**`?

src/core/globals.h
```diff
@@ -131,6 +133,11 @@ CGlobalVars *getGlobalVars();
 namespace modules {
 class CModule;
 
+void Initialize();
+CModule* GetModuleByName(std::string name);
+
+extern std::unordered_map<std::string, std::unique_ptr<CModule>> moduleList;
```
laper32, line 139: see comments from global.h

laper32 (commented):
ok

roflmuffin:
Testing this on Linux, the TestPlugin SwitchTeam hook no longer works, because it doesn't appear to be able to find the module:

`[01:09:54.385] CSSharp: Cannot find module /home/container/game/csgo/bin/linuxsteamrt64/libserver.so.`

roflmuffin (approved):
Tested on Linux

roflmuffin:
@Nukoooo Ready to merge?

Nukoooo:
I can't really think of anything to add/improve right now, so yeah please merge :)

Issue #258, opened 2024-01-14
"No frameworks were found" error on latest build 146


Tremolo4:
Hiya, just installed a fresh cs2 server with this mod for the first time.
Getting the error below (sorry for the console garbage characters).
I've then tried build 144 instead and it says `CSSharp: CounterStrikeSharp.API Loaded Successfully`, seems to work.
Guess something broke on the latest build.

```
[32m[21:36:26.993] CSSharp: Initializing[m
[32m[21:36:26.993] CSSharp: CoreConfig loaded.[m
[32m[21:36:27.012] CSSharp: Globals loaded.[m
[32m[21:36:27.038] CSSharp: Loading .NET runtime...[m
[32m[21:36:27.038] CSSharp: Loading hostfxr from /home/cs2/serverfiles/game/csgo/addons/counterstrikesharp/dotnet/host/fxr/8.0.1/libhostfxr.so[m
[32m[21:36:27.039] CSSharp: .NET Runtime Initialised.[m
[32m[21:36:27.039] CSSharp: Loading CSS API, Runtime Config: /home/cs2/serverfiles/game/csgo/addons/counterstrikesharp/api/CounterStrikeSharp.API.runtimeconfig.json[m
You must install or update .NET to run this application.

App: /home/cs2/serverfiles/game/bin/linuxsteamrt64/cs2
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '8.0.1' (x64)
.NET location: /home/cs2/serverfiles/game/csgo/addons/counterstrikesharp/dotnet

No frameworks were found.

Learn more:
https://aka.ms/dotnet/app-launch-failed

To install missing framework, download:
https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=8.0.1&arch=x64&rid=linux-x64&os=debian.12
[1m[41m[21:36:27.039] CSSharp: Init failed: -7fff7f6a[m
[31m[1m[21:36:27.039] CSSharp: Failed to load CSS API.[m
[31m[1m[21:36:27.039] CSSharp: Failed to initialize .NET runtime[m
[32m[21:36:27.039] CSSharp: Hooks added.[m
L 01/14/2024 - 21:36:27: [META] Loaded 1 plugin.
```

M1kep:
It seems the packaged dotnet runtime has some files in the wrong places.

[...]

PR #416, opened 2024-04-13
Feature/localization rework

Dliix66:
Added some more methods to simplify localization out of a command's context


roflmuffin:
Looks okay to me, might be worth adding a XML doc comment to the methods and updating the example plugin?

Dliix66:
I can do that 👍

frederikstonge:
Sine IStringLocalizer can be injected in other classes, would it not be better to create extension methods on IStringLocalizer?

Dliix66:
@roflmuffin I updated the example plugin and this should be available in the docs too.

roflmuffin:
I've added these as extension methods in build #258, which I think is cleaner/more extensible than amending the base class in this PR.

PR #418, opened 2024-04-14
The fix of getplayers

schwarper:
Ignoring ghost players who are not in the game, based on their SteamID, IP address, and connection status.


roflmuffin:
This should be fixed by #417

schwarper:
> This should be fixed by #417

Okay, thanks

PR #421, opened 2024-04-19
Addition of PlayerAim

schwarper:
- The feature of pulling the player in aim with `@aim` has been added.
- String values that should be char have been changed.
- Switch command modified

### DIFF
--- a/managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
+++ b/managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
@@ -22,6 +22,7 @@ public class Target
         { "@dead", TargetType.GroupDead },
         { "@!me", TargetType.GroupNotMe },
         { "@me", TargetType.PlayerMe },
+        { "@aim", TargetType.PlayerAim },
         { "@ct", TargetType.TeamCt },
         { "@t", TargetType.TeamT },
         { "@spec", TargetType.TeamSpec }
@@ -31,7 +32,7 @@ public class Target
     private static bool ConstTargetType(string target, out TargetType targetType)
     {
         targetType = TargetType.Invalid;
-        if (!target.StartsWith("@"))
+        if (!target.StartsWith('@'))
         {
             return false;
         }
@@ -45,7 +46,7 @@ public class Target
     {
         targetType = TargetType.Invalid;
         slug = null!;
-        if (!target.StartsWith("#"))
+        if (!target.StartsWith('#'))
         {
             return false;
         }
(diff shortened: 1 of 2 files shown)


schwarper:
I think it would be very nice if the GetClientAimTarget command can be pulled publicly from the Target library. That's why I didn't set it private.

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
```diff
@@ -124,4 +111,24 @@ public TargetResult GetTarget(CCSPlayerController? caller)
...
+    public CCSPlayerController? GetClientAimTarget(CCSPlayerController player)
+    {
+        CCSGameRules? GameRules = Utilities.FindAllEntitiesByDesignerName<CCSGameRulesProxy>("cs_gamerules").FirstOrDefault()?.GameRules;
+
+        if (GameRules == null)
+        {
+            return null;
+        }
+
+        VirtualFunctionWithReturn<IntPtr, IntPtr, IntPtr> findPickerEntity = new(GameRules.Handle, 28);
```
KillStr3aK, line 124: I dont really like this vtable offset hardcoded, should be in the gamedata instead

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
```diff
@@ -45,7 +46,7 @@ private static bool ConstTargetType(string target, out TargetType targetType)
     {
         targetType = TargetType.Invalid;
         slug = null!;
-        if (!target.StartsWith("#"))
+        if (!target.StartsWith('#'))
```
KillStr3aK, line 49: ?

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
```diff
@@ -31,7 +32,7 @@ public class Target
     private static bool ConstTargetType(string target, out TargetType targetType)
     {
         targetType = TargetType.Invalid;
-        if (!target.StartsWith("@"))
+        if (!target.StartsWith('@'))
```
KillStr3aK, line 35: ?

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
```diff
@@ -124,4 +111,24 @@ public TargetResult GetTarget(CCSPlayerController? caller)
         
         return new TargetResult() { Players = players };
     }
+
+    public CCSPlayerController? GetClientAimTarget(CCSPlayerController player)
+    {
+        CCSGameRules? GameRules = Utilities.FindAllEntitiesByDesignerName<CCSGameRulesProxy>("cs_gamerules").FirstOrDefault()?.GameRules;
```
KillStr3aK, line 117: this seems slow to me, might be good to find a way to cache gamerules to avoid calling it multiple times for no reason

KillStr3aK (commented):
rest LGTM

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
schwarper: > I dont really like this vtable offset hardcoded, should be in the gamedata instead

This code is a code shared in the CSS discord. Of course, it would be better to be developed and shared.

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
schwarper: > ?

Isn't it better to use char instead of string with StartsWith?

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
schwarper: > this seems slow to me, might be a good idea to find a way to cache gamerules to avoid calling it multiple times for no reason

Maybe, something like that?

```csharp
CCSGameRules GameRules { get; set; } = Utilities.FindAllEntitiesByDesignerName<CCSGameRulesProxy>("cs_gamerules").FirstOrDefault()!.GameRules!;

public CCSPlayerController? GetClientAimTarget(CCSPlayerController? player)
{
    if (player == null || player.IsValid)
    {
        return null;
    }

    VirtualFunctionWithReturn<IntPtr, IntPtr, IntPtr> findPickerEntity = new(GameRules.Handle, 28);
    CBaseEntity target = new(findPickerEntity.Invoke(GameRules.Handle, player.Handle));

    if (target.DesignerName is "player")
    {
        return target.As<CCSPlayerPawn>().OriginalController.Value;
    }

    return null;
}
```

KillStr3aK: this should and will fail when a new gamerules entity is created

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
KillStr3aK: could you add it to the gamedata tho?
also are you sure that 28 is the offset for both platforms?

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
KillStr3aK: there is a slight difference in the performance but thats it

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
```diff
@@ -12,6 +12,7 @@ public class Target
     private TargetType Type { get; }
     private string Raw { get; }
     private string Slug { get; }
+    private CCSGameRules GameRules { get; set; } = Utilities.FindAllEntitiesByDesignerName<CCSGameRulesProxy>("cs_gamerules").FirstOrDefault()!.GameRules!;
```
KillStr3aK, line 15: this gonna fail between sessions

PR #424, opened 2024-04-25
Make shared type loader less strict

KillStr3aK:
I propose this change as I don't think we should be this strict when we are loading shared types.
Currently when a shared type is loaded it is saved with assembly build version and plugins that depends on other shared type(s) cannot be loaded if the shared type version has changed, even if there are no breaking changes as the shared type just cannot be found at all (or version mismatch)

this pr changes this behaviour and it only looks for the shared type name and loads it, I believe developers should be aware of the changes of the shared type they depend on, just like how it works with the core api.


KillStr3aK:
(this has been tested and works as intended, also keep in mind that we can't even have the same assembly with different versions on the server)

roflmuffin (approved):

PR #432, opened 2024-04-30
feat: add "SteamAuth" core config option

roflmuffin:
Adds new "SteamAuth" core config option which controls how strict the "AuthorizedSteamID" property is with regard to steam connection.


AmbyCe (approved):

AmbyCe (approved):

crashzk (approved):

Mesharsky:
@roflmuffin Any update on that?
When is release gonna happen with that method.

roflmuffin:
I'm going to close this, as I'm not sure its the best idea; however we could probably do better to explain the difference between Authorized vs regular in docs. If there is an auth bug in CS# please repro it and raise it as an issue.

PR #433, opened 2024-04-30
AcceptInput changes

### LINKED ISSUE #238: Parameters don't work in CEntityInstance.AcceptInput
The function is incorrectly defined with a string for the input parameter when it should be a `variant_t`, see https://github.com/alliedmodders/hl2sdk/blob/cs2/public/variant.h

Yarukon:
(Might need to test on linux cuz i don't have linux enviroment)

Moved AcceptInput to NativeAPI, this change will fix #238
Added CEntitySystem_AddEntityIOEvent

### DIFF
--- a/src/scripting/natives/natives_entities.cpp
+++ b/src/scripting/natives/natives_entities.cpp
@@ -195,6 +195,8 @@ void UnhookEntityOutput(ScriptContext& script_context)
 
 void AcceptInput(ScriptContext& script_context)
 {
+    if (!CEntityInstance_AcceptInput) return;
+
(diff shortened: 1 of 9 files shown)


src/scripting/natives/natives_entities.cpp
```diff
@@ -195,6 +195,8 @@ void UnhookEntityOutput(ScriptContext& script_context)
 
 void AcceptInput(ScriptContext& script_context)
 {
+    if (!CEntityInstance_AcceptInput) return;
```
roflmuffin, line 198: I think we can hard-fail at this point using `script_context.ThrowNativeError()`

roflmuffin (approved):

PR #440, opened 2024-05-03
[Draft] EntityKeyValues implementation

Yarukon:
As the title said
(I hope there's a better way to implement this)


managed/CounterStrikeSharp.API/Modules/Utils/CEntityKeyValues.cs
```diff
@@ -0,0 +1,229 @@
...
+            public KeyValueContainer(KeyValuesType type, object value)
+            {
+                this.type = type;
+                this.value = value;
+            }
+
+            public KeyValuesType GetContainerType() => type;
+            public T Get<T>() => (T)value;
+
+            // Shut visual studio up
```
KillStr3aK, line 183: my man ❤️

Yarukon:
seems like the game will crash if there's too many keyvalues, caused by freeing the string ptr
image

PR #461, opened 2024-05-24
[Draft] Signature Fixes after 2024-05-23 CS2 update

roflmuffin:


zonical:
Tested on Linux (25/05). Server loads correctly and is able to change maps successfully. Haven't tested every single changed offset/signature but everything feels stable.

PR #479, opened 2024-06-05
Fix AddResource offset for Linux

Yarukon:
need those who use linux to test this to see if the crash still occur
(i dont have linux enviroment to test this rn)
how to test:
try precache something and then change map

### DIFF
--- a/configs/addons/counterstrikesharp/gamedata/gamedata.json
+++ b/configs/addons/counterstrikesharp/gamedata/gamedata.json
@@ -237,7 +237,7 @@
   "CEntityResourceManifest_AddResource": {
     "offsets": {
       "windows": 2,
-      "linux": 2
+      "linux": 1
     }
   }
 }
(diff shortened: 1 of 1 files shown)


Yarukon:
> I believe the offset should be 0 on linux and 2 on windows, this is because the first 3 functions in IEntityResourceManifest are all overloads of AddResource:
> 
> ```c++
> class IEntityResourceManifest
> {
> public:
> 	virtual void AddResource(const char*) = 0;
> 	virtual void AddResource(const char*, void*) = 0;
> 	virtual void AddResource(const char*, void*, void*, void*) = 0;
> 	virtual void unk_04() = 0;
> 	virtual void unk_05() = 0;
> 	virtual void unk_06() = 0;
> 	virtual void unk_07() = 0;
> 	virtual void unk_08() = 0;
> 	virtual void unk_09() = 0;
> 	virtual void unk_10() = 0;
> };
> ```
> 
> So what you want to use is the first one, or offset 0. But MSVC flips overload ordering so it becomes the third function, or offset 2.

changed to 0

xLeviNx:
Tested. This can be merged. 

@roflmuffin

roflmuffin (approved):

PR #556, opened 2024-08-22
Update DynoHook

Yarukon:
as the title said
TODO: need to handle the string differently, otherwise native side will get gibberish

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,9 +17,6 @@
 [submodule "libraries/DynoHook"]
 	path = libraries/DynoHook
 	url = https://github.com/qubka/DynoHook
-[submodule "libraries/asmjit"]
-	path = libraries/asmjit
-	url = https://github.com/asmjit/asmjit
 [submodule "libraries/Protobufs"]
 	path = libraries/Protobufs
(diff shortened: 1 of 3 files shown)


roflmuffin:
Testing this branch on Linux I seem to get some crashes when hooking functions that previously worked (usually string related methods), I'll see if I can get a repro case

Nianmou:
Nowadays, many functions of dynobooks do not work in Linux

oylsister:
> Nowadays, many functions of dynobooks do not work in Linux

So you tried out the latest dynohook? is there any alternative that we can use?

darkerz7:
Are you not going to update dynohook? `CCSPlayer_WeaponServices_CanUse` function on linux only works with the updated dynohook

Yarukon:
> Are you not going to update dynohook? `CCSPlayer_WeaponServices_CanUse` function on linux only works with the updated dynohook (maybe the latest version without crash, but with hook canuse -> https://github.com/qubka/DynoHook/tree/4742bfa04ca5b8f8ca70718e0088f715b92a57b6)

i've already ditched this method for function hooking, so yes, i dont have time.

PR #559, opened 2024-08-25
Added UTIL_SayText2Filter

schwarper:
Works for windows & linux finally
Thanks to @Yarukon


roflmuffin:
Will be solved by #435

PR #563, opened 2024-08-29
feat: add replicate convar

roflmuffin:


Cruze03:
@roflmuffin Can we add `Server.PrintToChatAll` and `player.PrintToChat` also in the this? Like using protobuf to do this would be better instead of signature functions.

KillStr3aK:
I believe we could also expose `CServerSideClient` into managed?

KillStr3aK:
> I believe we could also expose `CServerSideClient` into managed?

code snippets, offsets, etc: https://discord.com/channels/1160907911501991946/1231712355784851497 (also scroll down)
sending it just in case

Cruze03:
@roflmuffin Any update here :(? Would be nice if you could merge this after adding `CServerSideClient` into managed and send protobuff message for PrintToChats instead of signature funcs

M-archand:
Really looking forward to when this feature is added 🙂

M-archand:
Pls add to css rofl :)

PR #569, opened 2024-09-01
Implement CVariant functions

zonical:
~~**Very work in progress - working on it in small bits at a time**~~

This PR aims to implement functions to `CVariant` which accesses data from `variant_t` on the C++ side. Useful for functions that hook into the Entity IO system. ~~Unsure if a Set function is needed - but it can probably be done.~~


zonical:
I've also added a function to get a string from a `CUtlSymbolLarge` class for plugin authors that wish to hook into `AcceptInput` without dealing with memory shenanigans on the C# side. Does not appear to have an easy way to modify it (from my little knowledge of hl2sdk) so it will remain as read-only.

Yarukon:
you can just use fieldtype_t to replace VariantFieldTypes, fieldtype_t is from the schema

KillStr3aK:
LGTM!

> Does not appear to have an easy way to modify it (from my little knowledge of hl2sdk) so it will remain as read-only.
Maybe modify it through a native call?

zonical:
Finally had the time to work on this again. Added a Set() function with corresponding native functions. Currently untested (and I can't test at the moment - so I would be grateful if someone could!) 

@Yarukon - I've also used the generated field as you suggested. Thanks for the suggestion!

src/scripting/natives/natives_cutil.yaml
```diff
@@ -0,0 +1 @@
+GET_STRING_FROM_SYMBOL_LARGE: pPointer:pointer -> string
```
KillStr3aK, line 1: instead of `pPointer` it should be `pointer`

managed/CounterStrikeSharp.API/CompatibilitySuppressions.xml
```diff
@@ -1,6 +1,102 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids -->
-<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+<Suppressions xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
KillStr3aK, line 3: I believe this is not intended?

src/scripting/natives/natives_cutil.yaml
zonical: Addressed!

managed/CounterStrikeSharp.API/Core/Model/CUtlSymbolLarge.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Reflection.Metadata;
```
KillStr3aK, line 1: not to be picky, just trying to keep things clean (we have messed up already haha)
is this referenced?

managed/CounterStrikeSharp.API/Core/Model/CVariant.cs
```diff
@@ -1,13 +1,56 @@
-﻿namespace CounterStrikeSharp.API.Core;
+﻿using System.Reflection.Metadata;
```
KillStr3aK, line 1: same question as above

src/scripting/natives/natives_cutil.cpp
```diff
@@ -0,0 +1,44 @@
...
+ *  along with CounterStrikeSharp.  If not, see <https://www.gnu.org/licenses/>. *
+ */
+
+#include <public/entity2/entitysystem.h>
+
+#include <ios>
+#include <sstream>
+
+#include "core/log.h"
+#include "core/memory.h"
```
KillStr3aK, line 23: I believe these are needless there?

src/scripting/natives/natives_cvariant.cpp
```diff
@@ -0,0 +1,175 @@
...
+ *  along with CounterStrikeSharp.  If not, see <https://www.gnu.org/licenses/>. *
+ */
+
+#include <public/entity2/entitysystem.h>
+
+#include <ios>
+#include <sstream>
+
+#include "core/log.h"
+#include "core/memory.h"
```
KillStr3aK, line 23: I believe these are needless there aswell

KillStr3aK:
I've left probably my last reviews on this, good job!

zonical:
@KillStr3aK Your cleanup suggestions have been made!

KillStr3aK:
LGTM!

KillStr3aK (approved):
LGTM

KillStr3aK (approved):

KillStr3aK (approved):

darkerz7:
will this be added to the main branch?

zonical:
> will this be added to the main branch?

I hope so. I've been using a local version of CounterStrikeSharp with this feature for a while, and I've not encountered any massive issues. cc: @roflmuffin for review?

zonical:
> > will this be added to the main branch?
> 
> I hope so. I've been using a local version of CounterStrikeSharp with this feature for a while, and I've not encountered any massive issues. cc: @roflmuffin for review?

I did actually run into a crash or two, so I might investigate this code a bit further.

roflmuffin:
Hey @zonical, the changes look good to me, do you anticipate that your latest commit adds more safety to prevent the crashes? Do you need some more time to test

zonical:
Tried doing a massive rebase with latest code (this branch went extremely stale) and it fucked everything. New PR with a clean slate coming shortly!

PR #584, opened 2024-10-03
fix: armory update broken signatures and offsets

KillStr3aK:
~~We still miss `StateChanged` and `CCSPlayer_WeaponServices_CanUse` is unreliable on windows, linux is fine according to gdc~~

~~No idea if there is any broken offset but I guess so~~
fixed offsets:
- [CBaseEntity_Teleport](https://github.com/roflmuffin/CounterStrikeSharp/pull/584/commits/56059e852b2d25e01f392b124933e2b704f3af4c)
- [CCSPlayerController_ChangeTeam](https://github.com/roflmuffin/CounterStrikeSharp/pull/584/commits/f70cc33fcfccca116a9bd635ddcb3563346c98e2)
- [CBasePlayerPawn_CommitSuicide](https://github.com/roflmuffin/CounterStrikeSharp/pull/584/commits/a66c94902ed33e25b31dc98a3a431cc724f6e075)
- [CCSPlayer_ItemServices_DropActivePlayerWeapon](https://github.com/roflmuffin/CounterStrikeSharp/pull/584/commits/8cec48467dfc90da939dbdabda795a54e920b0c1)
- [CCSPlayer_ItemServices_GiveNamedItem](https://github.com/roflmuffin/CounterStrikeSharp/pull/584/commits/fadd9d7a66393698016712613150c20cfc508231)
- [CCSPlayerController_Respawn](https://github.com/roflmuffin/CounterStrikeSharp/pull/584/commits/ca010023c772d204f53f913866c0467a0955c8d1)
- [CCSPlayer_ItemServices_RemoveWeapons](https://github.com/roflmuffin/CounterStrikeSharp/pull/584/commits/bf1142d8a1d2e8d7949fbc66e7349fa7d7a53b46)

_most of the signatures are from stefanx_

PR #585, opened 2024-10-03
update cs2 offset

number201724:


KillStr3aK:
duplicate of #584 ?

number201724:
It looks like so.

PR #608, opened 2024-10-06
Implemented entity transmit feature

KillStr3aK:
With this implementation, developers are able to control the entity transmit per player

# Examples

These are simple examples, but it can be used for way more complex things.

```csharp
// hide every door for everyone as an example
RegisterListener<Listeners.CheckTransmit>((CCheckTransmitInfoList infoList) =>
{
    IEnumerable<CPropDoorRotating> doors = Utilities.FindAllEntitiesByDesignerName<CPropDoorRotating>("prop_door_rotating");

    if (!doors.Any())
        return;

    foreach ((CCheckTransmitInfo info, CCSPlayerController? player) in infoList)
    {
        if (player == null)
            continue;

        foreach (CPropDoorRotating door in doors)
        {
            info.TransmitEntities.Remove(door);
        }
    }
});
```

image


managed/CounterStrikeSharp.API/Core/Model/CFixedBitVecBase.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void Clear(int bitNum)
+        {
+            if (!(bitNum >= 0 && bitNum < MAX_EDICTS))
+                return;
+
+            uint* pInt = m_Ints + BitVec_Int(bitNum);
+            *pInt &= ~(uint)BitVec_Bit(bitNum);
+        }
+
+        public void Clear(uint BitNum)
```
KillStr3aK, line 41: This is only for saving people from doing a cast

managed/CounterStrikeSharp.API/Core/Model/CFixedBitVecBase.cs
```diff
@@ -0,0 +1,64 @@
...
+        public bool Contains(int bitNum)
+        {
+            if (!(bitNum >= 0 && bitNum < MAX_EDICTS))
+                return false;
+
+            uint* pInt = m_Ints + BitVec_Int(bitNum);
+            return (*pInt & (uint)BitVec_Bit(bitNum)) != 0;
+        }
+
+        public bool Contains(uint BitNum)
```
KillStr3aK, line 55: This is only for saving people from doing a cast

managed/CounterStrikeSharp.sln
```diff
@@ -1,5 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.12.35309.182
+MinimumVisualStudioVersion = 10.0.40219.1
```
KillStr3aK, line 5: not sure about these, probably was autogenerated when saving

managed/CounterStrikeSharp.sln
```diff
@@ -141,10 +153,14 @@ Global
...
-		{4E5289B5-E81D-421C-B340-B98B6FFE09D1} = {7DF99C35-881D-4FF2-B1C9-246BD3DECB9A}
 		{1309954E-FAF7-47A5-9FF9-C7263B33E4E3} = {7DF99C35-881D-4FF2-B1C9-246BD3DECB9A}
+		{4E5289B5-E81D-421C-B340-B98B6FFE09D1} = {7DF99C35-881D-4FF2-B1C9-246BD3DECB9A}
 		{A37676EA-CF2F-424D-85A1-C359D07A679D} = {7DF99C35-881D-4FF2-B1C9-246BD3DECB9A}
 		{76AD7BB0-A096-4336-83E2-B32CAE4E9933} = {7DF99C35-881D-4FF2-B1C9-246BD3DECB9A}
 		{A14029BA-CADE-4F25-ADC5-48CF14332F61} = {7DF99C35-881D-4FF2-B1C9-246BD3DECB9A}
+		{854B06B5-0E7B-438A-BA51-3F299557F884} = {7DF99C35-881D-4FF2-B1C9-246BD3DECB9A}
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {069C4CD4-BACA-446A-A6B8-0194E4F75355}
```
KillStr3aK, line 164: not sure about these, probably was autogenerated when saving

managed/CounterStrikeSharp.API/Core/Model/CFixedBitVecBase.cs
```diff
@@ -0,0 +1,64 @@
...
+    public unsafe struct CFixedBitVecBase
+    {
+        private const int LOG2_BITS_PER_INT = 5;
+        private const int MAX_EDICT_BITS = 14;
+        private const int BITS_PER_INT = 32;
+        private const int MAX_EDICTS = 1 << MAX_EDICT_BITS;
+
+        private readonly uint* m_Ints;
+
+        public void Clear(int bitNum)
```
KillStr3aK, line 32: Maybe we can rename this to `Remove`?

managed/CounterStrikeSharp.API/Core/Model/CFixedBitVecBase.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Runtime.InteropServices;
+
+namespace CounterStrikeSharp.API.Core
+{
+    // credits: qstage
+    [StructLayout(LayoutKind.Sequential)]
+    public unsafe struct CFixedBitVecBase
+    {
+        private const int LOG2_BITS_PER_INT = 5;
+        private const int MAX_EDICT_BITS = 14;
```
roflmuffin, line 26: Might be able to reuse the values from utilities: https://github.com/roflmuffin/CounterStrikeSharp/blob/42dd270b7852c7d866a8bdf835524fc0bd27a787/managed/CounterStrikeSharp.API/Utilities.cs#L35

KillStr3aK: yes we can re use the values!

managed/CounterStrikeSharp.API/Core/Model/CFixedBitVecBase.cs
Dliix66: Agreed :)

KillStr3aK: In general `Clear` is the right name for that method since it clears the bits but since we are in C# it might be preferable to name it like `Remove`

managed/CounterStrikeSharp.API/Core/Model/CCheckTransmitInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace CounterStrikeSharp.API.Core
+{
+    [StructLayout(LayoutKind.Sequential)]
+    public struct CCheckTransmitInfo
+    {
+        /// <summary>
+        /// Entity n is already marked for transmission
+        /// </summary>
+        public CFixedBitVecBase TransmitEntity;
```
KillStr3aK, line 27: Opinions on naming this `TransmitEntities`?

aquevadis: Is in not better, `TransmittedEntities`, as they are `marked already` per initial summary, so it becomes `TransmittedEntities.Remove(entityIndex)`?

KillStr3aK: They are not transmitted yet! they are just marked for transmission, thats why I'm thinking about `TransmitEntities` (it sounds like 'list of entities for transmit')

roflmuffin:
Overall the PR looks great!😍 I am wondering though if we can possibly improve the DX of iterating over the transmit list. Maybe something like this?

```csharp
RegisterListener<Listeners.CheckTransmit>((CCheckTransmitInfoList infoList, int infoCount) =>
{
    // Get the list of the currently available doors (prop_door_rotating)
    IEnumerable<CPropDoorRotating> doors = Utilities.FindAllEntitiesByDesignerName<CPropDoorRotating>("prop_door_rotating");

    // Do nothing if there is none.
    if (!doors.Any())
        return;

    foreach(var (CFixedBitVecBase transmitEntities, CCSPlayerController? player) in infoList) 
    {
        if (player == null || !ShouldSeeDoors.ContainsKey(player.Slot)) 
            continue;
        
        foreach (CPropDoorRotating door in doors)
        {
            transmitEntities.Remove(door);
        }   
    }
});
```

The idea being that if we can provide access to an `IEnumerable` (or even `ICollection`) ourselves, we can handle the bounds checking and prevent issues with that, as well as simplifying the access to the entities list & player.

KillStr3aK:
> Overall the PR looks great!😍 I am wondering though if we can possibly improve the DX of iterating over the transmit list. Maybe something like this (pseudo-ish code)?
> 
> ```cs
> RegisterListener<Listeners.CheckTransmit>((CCheckTransmitInfoList infoList, int infoCount) =>
> {
>     // Get the list of the currently available doors (prop_door_rotating)
>     IEnumerable<CPropDoorRotating> doors = Utilities.FindAllEntitiesByDesignerName<CPropDoorRotating>("prop_door_rotating");
> 
>     // Do nothing if there is none.
>     if (!doors.Any())
>         return;
> 
>     foreach(var (CFixedBitVecBase transmitEntities, CCSPlayerController? player) in infoList) 
>     {
>         if (player == null || !ShouldSeeDoors.ContainsKey(player.Slot)) 
>             continue;
>         
>         foreach (CPropDoorRotating door in doors)
>         {
>             transmitEntities.Remove(door);
>         }   
>     }
> });
> ```
> 
> The idea being that if we can provide access to an `IEnumerable` ourselves (or preferably `IReadOnlyList<T>` so that we can maintan index access), we can handle the bounds checking and prevent issues with that, as well as simplifying the access to the entities list & player.
> 
> Example enumerator implementation on `CCheckTransmitInfoList`. But we have to somehow pass the length of the transmit list through from native to managed, which looks hard with the current implementation simply casting from a pointer:
> 
> ```cs
> public sealed class CCheckTransmitInfoList : NativeObject,
>         IReadOnlyList<(CFixedBitVecBase TransmitEntities, CCSPlayerController? Player)>
>     {
>         private int CheckTransmitPlayerSlotOffset = GameData.GetOffset("CheckTransmitPlayerSlot");
> 
>         private unsafe nint* Inner => (nint*)base.Handle;
> 
>         public unsafe CCheckTransmitInfoList(IntPtr pointer) : base(pointer)
>         {
>         }
> 
>         public int Count { get; }
> 
> 
>         private unsafe (CCheckTransmitInfo, int) Get(int index)
>         {
>             return (Marshal.PtrToStructure<CCheckTransmitInfo>(this.Inner[index]),
>                 (int)(*(byte*)(this.Inner[index] + CheckTransmitPlayerSlotOffset)));
>         }
> 
>         public IEnumerator<(CFixedBitVecBase TransmitEntities, CCSPlayerController? Player)> GetEnumerator()
>         {
>             for (int i = 0; i < this.Count; i++)
>             {
>                 yield return this[i];
>             }
>         }
> 
>         IEnumerator IEnumerable.GetEnumerator()
>         {
>             return GetEnumerator();
>         }
> 
>         /// <summary>
>         /// Get transmit info for the given index.
>         /// </summary>
>         /// <param name="index">Index of the info you want to retrieve from the list, should be between 0 and 'infoCount' - 1</param>
>         /// <returns></returns>
>         public (CFixedBitVecBase TransmitEntities, CCSPlayerController? Player) this[int index]
>         {
>             get
>             {
>                 // Ideally throw here if out of range
>                 var (transmit, slot) = this.Get(index);
>                 CCSPlayerController? player = Utilities.GetPlayerFromSlot(slot);
>                 return (transmit.TransmitEntities, player);
>             }
>         }
>     }
> ```

Great idea! Will take a look on it

Dliix66 (approved):
Working properly on a test environment. Might be worth adding a warning if hiding a pawn, as killing a hidden pawn crashes the client.

KillStr3aK:
> Working properly on a test environment! Might be worth adding a warning if hiding a pawn, as killing a hidden pawn crashes the client.
> 
> cc @KillStr3aK

Well, currently its only noted in the example plugin but devs should be aware of this issue

roflmuffin (approved):

PR #613, opened 2024-10-07
[DRAFT] Featuring `MemoryManager` to resolve issues related to native memory leaking

KillStr3aK:
Hey there, I propose this as a 'demo' currently. It works but not sure if this is the right implementation for this on the managed side, so I'm open for any ideas or extensions.

Currently every instance of `Vector`/`QAngle`/`Angle` creates a new pointer in unmanaged that is stored but never released, therefore it leaks memory which has been known since they were implemented.

I believe that the current implementation that involves `IDisposableMemory`/`DisposableMemory` is a breaking change, however we can drop this if we reuse the same code in every class that wraps around an internal pointer (in my opinion, I'd go with breaking changes as we are supposed to have more classes like `Vector2D`, `Vector4D`, `Quaternion` and so on.)

I believe we should have natives for `MemAlloc` and rework a bit of the code in this branch (we would not need to have `NativeAPI.AngleDelete`/`NativeAPI.VectorDelete` and `managed_vectors`, `managed_angles` since we could just release the ptr that is stored in `NativeObject`)

I have no performance benchmarks in production with forcing `GC` yet, also should change `CoreConfig.MemoryManagerInterval` to a viable interval, it was only used for tests.

Since these types become `IDisposable` this syntax is now allowed:

```csharp
using (Vector vector = new Vector(150.0f, 150.0f, 150.0f))
{
    // work with the vector here
}

// `vector` instance is now released automatically without GC
```

Instances that are not used like this will be collected by the garbage collector and the finalizer will call the responsible native for releasing the internal pointer

Releasing `Vector` instances:
image

```csharp
// Instances like these are not supposed to be collected, and they won't be as long as its stored/not changed. Otherwise the GC will clean it up
public Vector ThisShouldNotBeGarbageCollected = new Vector();
```
image

I will continue working on this, and once we get this one done we can implement `CEntityKeyValues`


KillStr3aK:
also features class `MemAlloc`

```csharp
public static class MemAlloc
{
    /// <summary>
    /// Indirect allocation using 'MemAlloc'
    /// </summary>
    /// <param name="size"></param>
    /// <returns></returns>
    public static nint Allocate(int size)
    {
        return NativeAPI.MemAllocAllocate(size);
    }

    /// <summary>
    /// Indirect allocation using 'MemAlloc'
    /// </summary>
    /// <param name="size"></param>
    /// <param name="align"></param>
    /// <returns></returns>
    public static nint AllocateAligned(int size, int align)
    {
        return NativeAPI.MemAllocAllocateAligned(size, align);
    }

    public static nint ReallocateAligned(nint pointer, int size, int align)
    {
        return NativeAPI.MemAllocReallocateAligned(pointer, size, align);
    }

    public static int GetSizeAligned(nint pointer)
    {
        return NativeAPI.MemAllocGetSizeAligned(pointer);
    }

    /// <summary>
    /// Release pointer using 'MemAlloc'
    /// </summary>
    /// <param name="ptr"></param>
    public static void Free(nint ptr)
    {
        NativeAPI.MemAllocFreePointer(ptr);
    }

    /// <summary>
    /// Release pointer using 'MemAlloc'
    /// </summary>
    /// <param name="ptr"></param>
    public static void FreeAligned(nint ptr)
    {
        NativeAPI.MemAllocFreePointerAligned(ptr);
    }
}
```

managed/CounterStrikeSharp.API/CounterStrikeSharp.API.csproj
```diff
@@ -23,47 +23,43 @@
...
     </PropertyGroup>
     <Target Name="SetSourceRevisionId" BeforeTargets="InitializeSourceControlInformation">
-        <Exec
-                Command="git describe --long --always --exclude=* --abbrev=7"
-                ConsoleToMSBuild="True"
-                IgnoreExitCode="False"
-        >
-            <Output PropertyName="SourceRevisionId" TaskParameter="ConsoleOutput"/>
+        <Exec Command="git describe --long --always --exclude=* --abbrev=7" ConsoleToMSBuild="True" IgnoreExitCode="False">
+            <Output PropertyName="SourceRevisionId" TaskParameter="ConsoleOutput" />
```
KillStr3aK, line 62: IDE probably auto formatted it, I'm not sticking to this change haha, can be reversed

managed/CounterStrikeSharp.API/Core/Memory/MemoryManager.cs
```diff
@@ -0,0 +1,318 @@
...
+                {
+                    _logger.LogInformation("Released {0} leaking memory resources in {1}ms ({2} remains)", LastReleased, (LastUpdated - startTime).TotalMilliseconds, CurrentResources);
+                } else
+                {
+                    Thread.Sleep(CoreConfig.MemoryManagerInterval);
+                }
+            }
+        }
+    }
+}
```
KillStr3aK, line 318: I'm concerned about this, not sure if this is the right way to do this?

managed/CounterStrikeSharp.API/Core/Memory/MemoryManager.cs
```diff
@@ -0,0 +1,318 @@
...
+
+                _logger.LogInformation("Starting service...");
+
+                try
+                {
+                    _thread.Start();
+                } catch (ThreadStateException)
+                {
+                    _thread = new Thread(BackgroundThread);
+                    _thread.Start();
```
KillStr3aK, line 208: this handles thread creation, if it was terminated earlier this exception is thrown

managed/CounterStrikeSharp.API/Core/Memory/MemoryManager.cs
```diff
@@ -0,0 +1,318 @@
...
+        private void PrintStatistics(CommandInfo info)
+        {
+            info.ReplyToCommand("Memory Manager Statistics:\n" +
+                                      $"State: {State}\n" +
+                                      $"Total Released: {TotalReleased}\n" +
+                                      $"Last Released: {LastReleased}\n" +
+                                      $"Current Resources: {CurrentResources}\n" +
+                                      $"Last Updated: {LastUpdated.ToString("yyyy.MM.dd hh:mm:ss tt")} (UTC)\n" +
+                                      $"Heap Memory Usage: ~{TotalMemoryMB:F5} MB ({TotalMemory} bytes)\n" +
+                                      $"Total Allocated Bytes: ~{TotalAllocatedMB:F5} MB ({TotalAllocated} bytes)");
```
KillStr3aK, line 181: not sure if we should call those GC methods multiple times, I'm down for any change

KillStr3aK:
In general the `MemoryManager` just enforces the `GC` to do its job so stuff that can be collected **will be** collected imminently, even if we drop this stuff the junk/leaks will be collected, we just don't know when.

managed/CounterStrikeSharp.API/Core/Memory/MemoryManager.cs
```diff
@@ -0,0 +1,320 @@
...
+
+        public MemoryManager(ILogger<MemoryManager> logger, ICommandManager commandManager)
+        {
+            _logger = logger;
+            _commandManager = commandManager;
+
+            _thread = new Thread(BackgroundThread);
+        }
+
+        [RequiresPermissions("@css/generic")]
```
KillStr3aK, line 112: afaik we no longer use the `root` flag?

oylsister:
Is this pretty much done or still have something to work on?

KillStr3aK:
> Is this pretty much done or still have something to work on?

I believe we should look for another way of doing this as the cleanup comes with overhead, but I'd assume that any change that would fix this issue would be a breaking one. I'd prefer a dev branch with breaking changes to give plugin developers more time to adjust their codebase than the current situation

2vg:
I experimented with this in a private fork.
Releasing vectors works well, but with something like BulletTracer, there might be overhead in releasing them after 64 people shoot at once.
Since I use Satori for GC, it doesn't bother me much in my environment but I thought of something like a `DisposableBag` to reduce overhead, but I haven't tried it with a full player yet.
With this bag, I think it should be possible to solve the problem by releasing `Span<T>` every frame as well... (On the Managed side, I just mark them, and release them all or little by little on the GameFrame hook...)
Another idea for compatibility, I'm considering an implicit operator to minimize it.
It won't be useful if we explicitly specify `Span<T>`, but if we declare it as a `var`, we should be able to convert it implicitly. :eyes:
Anyway, thanks for the great work!

KillStr3aK:
> I experimented with this in a private fork. Releasing vectors works well, but with something like BulletTracer, there might be overhead in releasing them after 64 people shoot at once. Since I use Satori for GC, it doesn't bother me much in my environment but I thought of something like a `DisposableBag` to reduce overhead, but I haven't tried it with a full player yet. With this bag, I think it should be possible to solve the problem by releasing `Span<T>` every frame as well... (On the Managed side, I just mark them, and release them all or little by little on the GameFrame hook...) Another idea for compatibility, I'm considering an implicit operator to minimize it. It won't be useful if we explicitly specify `Span<T>`, but if we declare it as a `var`, we should be able to convert it implicitly. 👀 Anyway, thanks for the great work!

pretty sure we would need to get rid of the native call overhead to do this, thanks for the info!

PR #617, opened 2024-10-09
Extend `Dynamic`

KillStr3aK:


managed/CounterStrikeSharp.API/Modules/Memory/DynamicFunctions/BaseMemoryFunction.cs
```diff
@@ -45,6 +47,30 @@ private static IntPtr CreateValveFunctionBySignature(string signature, string bi
...
+            {
+                function = NativeAPI.CreateVirtualFunction(objectPtr, offset,
+                    argumentTypes.Length, (int)returnType, argumentTypes.Cast<object>().ToArray());
+                createdFunctions[offset] = function;
+            } catch (Exception)
+            {
+            }
+        }
+
+        return function;
```
KillStr3aK, line 73: might need a better way? if we pass another entity instance (ptr) with the same offset it will probably try to hook the same function again and fail? (untested currently, will answer here when I can test it)

Prefix:
@roflmuffin +1 on this

KillStr3aK:
> @roflmuffin +1 on this

my second review has to be resolved before this pr is ready

Prefix:
> > @roflmuffin +1 on this
> 
> my second review has to be resolved before this pr is ready

what is second pr? can community help you somehow?

KillStr3aK:
> > > @roflmuffin +1 on this
> 
> > 
> 
> > my second review has to be resolved before this pr is ready
> 
> 
> 
> what is second pr? can community help you somehow?

https://github.com/roflmuffin/CounterStrikeSharp/pull/617#discussion_r1793274716

this is the second review

managed/CounterStrikeSharp.API/Modules/Memory/DynamicFunctions/BaseMemoryFunction.cs
roflmuffin: We could possibly ask for a "name" to disambiguate what the Virtual Function represents? For instance 
```
if (OnStartTouch == null) (OnStartTouch = new VirtualFunctionVoid<CBaseEntity, CBaseEntity>(SoccerBall, "CBaseEntityStartTouch", GameData.GetOffset("StartTouch"))).Hook(HookOnStartTouch, HookMode.Pre);
```

Another alternative would be to construct a key using the passed type argument, so it would automatically become `CBaseEntity_342` for instance. This might be my preferred way. Unsure of the interaction if someone hooks a `CBaseEntity` and a `CCSPlayerPawn` with the same offset.

KillStr3aK: > We could possibly ask for a "name" to disambiguate what the Virtual Function represents? For instance
> 
> ```
> if (OnStartTouch == null) (OnStartTouch = new VirtualFunctionVoid<CBaseEntity, CBaseEntity>(SoccerBall, "CBaseEntityStartTouch", GameData.GetOffset("StartTouch"))).Hook(HookOnStartTouch, HookMode.Pre);
> ```
> 
> Another alternative would be to construct a key using the passed type argument, so it would automatically become `CBaseEntity_342` for instance. This might be my preferred way. Unsure of the interaction if someone hooks a `CBaseEntity` and a `CCSPlayerPawn` with the same offset.

The alternative method sounds good for me, will try it out today as I already have an idea

oylsister: Could this has to hook with entity every time? like If I want to hook on offset with DropWeapon I have to do every time when ``player_spawn``.

samyycX: Using the memory address of the function as key should be fine since they are unique

KillStr3aK: > Could this has to hook with entity every time? like If I want to hook on offset with DropWeapon I have to do hook ``CCSPlayer_WeaponServices`` every time when ``player_spawn``.

You would only have to set it up once, then every call to the vfunc is hooked, regardless of instance

samyycX: > might need a better way? if we pass another entity instance (ptr) with the same offset it will probably try to hook the same function again and fail? (untested currently, will answer here when I can test it)

Dynohook seems to have an internal same function check before hooking it, so i guess it wont cause problem

KillStr3aK:
virtual functions that has been created by an offset are now stored as `{typeof(TArg1).Name}_{OffsetNum}`, for e.g.:
`10:52:04 [INFO] (cssharp:Core) Creating function CBaseEntity_142`

`TArg1` is meant to represent the originating class of the function

> @roflmuffin Unsure of the interaction if someone hooks a CBaseEntity and a CCSPlayerPawn with the same offset.

I'd say that case would be a semantic error on its own already

darkerz7:
is there any plans to add this to the main branch?

KillStr3aK:
> is there any plans to add this to the main branch?

Hi, I plan to finish it in general, just need to find a better way to store these functions as the behaviour I've mentioned in my previous comment is not really viable in my opinion

KoNLiG:
Looking forward for it

roflmuffin:
This is fixed by #1143 and released

PR #627, opened 2024-10-13
Add GetGameframeTime to NativeAPI

Interesting-exe:
GetGameframeTime was [registered](https://github.com/roflmuffin/CounterStrikeSharp/blob/ad7f7bd365d7080060acc94539830a9066cc4633/src/scripting/natives/natives_engine.cpp#L248) but not used

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/API.cs
+++ b/managed/CounterStrikeSharp.API/Core/API.cs
@@ -337,6 +337,16 @@ namespace CounterStrikeSharp.API.Core
 			}
 		}
 
+        public static float GetGameframeTime(){
+			lock (ScriptContext.GlobalScriptContext.Lock) {
+			ScriptContext.GlobalScriptContext.Reset();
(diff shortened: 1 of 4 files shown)


managed/CounterStrikeSharp.API/Core/API.cs
```diff
@@ -337,6 +337,16 @@ public static int GetMaxClients(){
 			}
 		}
 
+        public static float GetGameframeTime(){
```
KillStr3aK, line 340: might be better as `GetGameFrameTime` (use `GET_GAME_FRAME_TIME` in the yaml)

KillStr3aK:
other than these its LGTM

managed/CounterStrikeSharp.API/Core/API.cs
Interesting-exe: Yeah I agree, I just used the name that was already in the natives cpp.

roflmuffin (approved):

PR #628, opened 2024-10-13
Added GetCSWeaponDataFromKey and CCSPlayer_ItemServices_CanAcquire

schwarper:
CanUse doesn't work on Linux.
CanAcquire or GetCSWeaponDataFromKey doesn't work on Windows.

Credit:
https://github.com/CS2Plugins/WeaponRestrict


managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctions.cs
```diff
@@ -10,9 +10,29 @@ namespace CounterStrikeSharp.API.Modules.Memory;
...
+        AlreadyPurchased,
+        ReachedGrenadeTypeLimit,
+        ReachedGrenadeTotalLimit,
+        NotAllowedByTeam,
+        NotAllowedByMap,
+        NotAllowedByMode,
+        NotAllowedForPurchase,
+        NotAllowedByProhibition,
+    };
+
```
KillStr3aK, line 33: this should not be in here. maybe consider `Modules/Utils/*`, create files named the same as the enum

```
Models/Utils/AcquireMethod.cs
Models/Utils/AcquireResult.cs
```

managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctions.cs
```diff
@@ -10,9 +10,29 @@ namespace CounterStrikeSharp.API.Modules.Memory;
...
     public static MemoryFunctionVoid<IntPtr, HudDestination, string, IntPtr, IntPtr, IntPtr, IntPtr> ClientPrintFunc =
-        new(
-            GameData.GetSignature("ClientPrint"));
+        new(GameData.GetSignature("ClientPrint"));
```
KillStr3aK, line 35: this should be reverted to keep the style format

managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctions.cs
```diff
@@ -68,7 +88,13 @@ public static class VirtualFunctions
...
     public static MemoryFunctionWithReturn<CCSPlayer_WeaponServices, CBasePlayerWeapon, bool> CCSPlayer_WeaponServices_CanUseFunc = new(GameData.GetSignature("CCSPlayer_WeaponServices_CanUse"));
     public static Func<CCSPlayer_WeaponServices, CBasePlayerWeapon, bool> CCSPlayer_WeaponServices_CanUse = CCSPlayer_WeaponServices_CanUseFunc.Invoke;
-    
+
+    public static MemoryFunctionWithReturn<int, string, CCSWeaponBaseVData> GetCSWeaponDataFromKeyFunc = new(GameData.GetSignature("GetCSWeaponDataFromKey"));
+    public static Func<int, string, CCSWeaponBaseVData> GetCSWeaponDataFromKey = GetCSWeaponDataFromKeyFunc.Invoke;
+
+    public static MemoryFunctionWithReturn<CCSPlayer_ItemServices, CEconItemView, AcquireMethod, NativeObject, AcquireResult> CCSPlayer_CanAcquireFunc = new(GameData.GetSignature("CCSPlayer_CanAcquire"));
+    public static Func<CCSPlayer_ItemServices, CEconItemView, AcquireMethod, NativeObject, AcquireResult> CCSPlayer_CanAcquire = CCSPlayer_CanAcquireFunc.Invoke;
+
```
KillStr3aK, line 97: LGTM, untested

KillStr3aK:
Consider adding the function itself to the responsible class to make it callable? (`Core/Model/*`) as an extension method

managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctions.cs
KillStr3aK: maybe prefix the variables with the correct name? `CCSPlayer` -> `CCSPlayerController` (if its from the controller ofc)

managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctions.cs
```diff
@@ -68,7 +88,13 @@ public static class VirtualFunctions
     
     public static MemoryFunctionWithReturn<CCSPlayer_WeaponServices, CBasePlayerWeapon, bool> CCSPlayer_WeaponServices_CanUseFunc = new(GameData.GetSignature("CCSPlayer_WeaponServices_CanUse"));
     public static Func<CCSPlayer_WeaponServices, CBasePlayerWeapon, bool> CCSPlayer_WeaponServices_CanUse = CCSPlayer_WeaponServices_CanUseFunc.Invoke;
-    
+
+    public static MemoryFunctionWithReturn<int, string, CCSWeaponBaseVData> GetCSWeaponDataFromKeyFunc = new(GameData.GetSignature("GetCSWeaponDataFromKey"));
+    public static Func<int, string, CCSWeaponBaseVData> GetCSWeaponDataFromKey = GetCSWeaponDataFromKeyFunc.Invoke;
```
KillStr3aK, line 93: isn't this the same as `GetVData`?

schwarper: Probably yes, but there might not be a way to get vdata from CanAcquire

managed/CounterStrikeSharp.API/Modules/Utils/CanAcquire.cs
```diff
@@ -0,0 +1,22 @@
...
+    AlreadyOwned,
+    AlreadyPurchased,
+    ReachedGrenadeTypeLimit,
+    ReachedGrenadeTotalLimit,
+    NotAllowedByTeam,
+    NotAllowedByMap,
+    NotAllowedByMode,
+    NotAllowedForPurchase,
+    NotAllowedByProhibition,
+};
```
KillStr3aK, line 22: instead of `CanAcquire.cs`, create `AcquireResult.cs` and `AcquireMethod.cs` with the correspondive values in them. also dont forget the license text at the top

managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctions.cs
```diff
@@ -68,7 +67,13 @@ public static class VirtualFunctions
...
+
+    public static MemoryFunctionWithReturn<CCSPlayer_ItemServices, CEconItemView, AcquireMethod, NativeObject, AcquireResult> CCSPlayer_CanAcquireFunc = new(GameData.GetSignature("CCSPlayer_CanAcquire"));
+    public static Func<CCSPlayer_ItemServices, CEconItemView, AcquireMethod, NativeObject, AcquireResult> CCSPlayer_CanAcquire = CCSPlayer_CanAcquireFunc.Invoke;
+
```
KillStr3aK, line 76: LGTM, however I wonder if would it be possible to wrap extension methods around these? so we would not have to do `VirtualFunctions.CCSPlayer_CanAcquireFunc(..)` but `itemServices.CanAcquire(..)`

also use proper gamedata names, please rename from `CCSPlayer_CanAcquire` to `CCSPlayer_ItemServices_CanAcquire` (same with the other one if you know the exact class name)

KillStr3aK (changes requested):

KillStr3aK (approved):
LGTM

KillStr3aK (approved):

KillStr3aK (approved):

roflmuffin:
Looks good, my only comment is about the usage of `NativeObject` in the raw hook type, as we normally use `IntPtr` for any unknown pointers in the call.

schwarper:
> Looks good, my only comment is about the usage of `NativeObject` in the raw hook type, as we normally use `IntPtr` for any unknown pointers in the call. If it is meant to represent an entity instance then we should probably use a `CEntityInstance`.

I replaced `NativeObject` with `IntPtr`. I think it's okay to use IntPtr there. As far as I tested, it's always zero.

managed/CounterStrikeSharp.API/Core/Model/CCSPlayer_ItemServices.cs
```diff
@@ -54,4 +56,9 @@ public void RemoveWeapons()
 
         return (T)Activator.CreateInstance(typeof(T), pointer)!;
     }
+
+    public AcquireResult CanAcquire(CEconItemView itemview, AcquireMethod method, IntPtr? nativeobject)
+    {
+        return VirtualFunctions.CCSPlayer_ItemServices_CanAcquireFunc.Invoke(this, itemview, method, nativeobject ?? 0);
+    }
```
roflmuffin, line 63:
```suggestion
    public AcquireResult CanAcquire(CEconItemView itemView, AcquireMethod method, IntPtr unknown = 0)
    {
        return VirtualFunctions.CCSPlayer_ItemServices_CanAcquireFunc.Invoke(this, itemView, method, unknown);
    }
```

schwarper: > 

That looks better.

roflmuffin (approved):

PR #639, opened 2024-10-20
Added TargetType.PlayerAim

schwarper:


roflmuffin:
Looks good, but we should add the `28` as an offset into the gamedata rather than hardcode it in code

schwarper:
> Looks good, but we should add the `28` as an offset into the gamedata rather than hardcode it in code

What sort of?

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
```diff
@@ -80,48 +83,36 @@ public Target(string target)
...
+            TargetType.GroupDead => player.PlayerPawn is { IsValid: true, Value.LifeState: (byte)LifeState_t.LIFE_DEAD or (byte)LifeState_t.LIFE_DYING },
+            TargetType.GroupNotMe => player.SteamID != caller?.SteamID,
+            TargetType.PlayerMe => player.SteamID == caller?.SteamID,
+            TargetType.IdUserid => player.UserId.ToString() == Slug,
+            TargetType.IdSteamEscaped => ((SteamID)player.SteamID).SteamId2 == Slug,
+            TargetType.IdSteam64 => ((SteamID)player.SteamID).SteamId64.ToString() == Slug,
+            TargetType.ExplicitName => player.PlayerName.Contains(Slug, StringComparison.OrdinalIgnoreCase),
+            TargetType.ImplicitName => player.PlayerName.Contains(Slug, StringComparison.OrdinalIgnoreCase),
+            _ => false
+        };
```
KillStr3aK, line 107: are we sure we want to switch this whole part to this syntax?

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
```diff
@@ -80,48 +83,36 @@ public Target(string target)
...
     }
 
     public TargetResult GetTarget(CCSPlayerController? caller)
     {
-        var players = Utilities.GetPlayers().Where(player => TargetPredicate(player, caller)).ToList();
-        
+        CCSGameRules? gameRules = Utilities.FindAllEntitiesByDesignerName<CCSGameRulesProxy>("cs_gamerules").FirstOrDefault()?.GameRules;
```
KillStr3aK, line 112: I believe we should not get gamerules each time here, might worth using the `EntityCache<T>` from discord as a built in type? not sure, just a first guess

managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs
schwarper: It looks better imo

schwarper:
> > > Looks good, but we should add the `28` as an offset into the gamedata rather than hardcode it in code
> > 
> > 
> > What sort of?
> 
> https://github.com/roflmuffin/CounterStrikeSharp/blob/a93f050fd99f03bde526ff172d14939f34198a45/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs#L34

What should be the offset's name?

```json
"CCSGameRules_FindPickerEntity": {
    "offsets": {
      "windows": 28,
      "linux": 28
    }
  },
  ```

File: managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs (hunk 2 above)
Comment (schwarper) [reply managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs:112]: Can you give me an example?

**Comment by KillStr3aK** [comment]
> > > > Looks good, but we should add the `28` as an offset into the gamedata rather than hardcode it in code
> > > 
> > > 
> > > What sort of?
> > 
> > 
> > https://github.com/roflmuffin/CounterStrikeSharp/blob/a93f050fd99f03bde526ff172d14939f34198a45/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs#L34
> 
> What should be the offset's name?
> 
> ```json
> "CCSGameRules_FindPickerEntity": {
>     "offsets": {
>       "windows": 28,
>       "linux": 28
>     }
>   },
> ```

that one seems okay to me, not sure about its "real" name but I'm down for this

File: managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs (hunk 2 above)
Comment (KillStr3aK) [reply managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs:112]: but with that, it would only get gamerules if its invalid, otherwise it would use the cached value instead of getting it each time that method is called which is more expensive than it should be

Comment (KillStr3aK) [reply managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs:112]: also you would still need to check for null/not valid

File: managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs (hunk 1 above)
Comment (KillStr3aK) [reply managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs:107]: > It looks better imo

well, thats preference

I wonder though, can we set the gamerules nullable and have `null` as default value? I know thats its a private method but maybe only get and pass gamerules if the target type requires it

File: managed/CounterStrikeSharp.API/Modules/Commands/Targeting/Target.cs [hunk 3]
```diff
@@ -83,11 +83,11 @@ public Target(string target)
         }
     }
 
-    private bool TargetPredicate(CCSPlayerController player, CCSPlayerController? caller, CCSGameRules gameRules)
+    private bool TargetPredicate(CCSPlayerController player, CCSPlayerController? caller, CCSGameRules? gameRules)
```
KillStr3aK, line 86: `CCSGameRules? gameRules = null`

schwarper: We send null gamerules already. Do we need it?

KillStr3aK (approved):
LGTM

configs/addons/counterstrikesharp/gamedata/gamedata.json
```diff
@@ -133,6 +133,12 @@
       "linux": "55 48 89 E5 41 57 41 56 41 55 41 54 49 89 FC 53 48 81 EC 88 00 00 00 48 8D 05 ? ? ? ?"
     }
   },
+  "CCSGameRules_FindPickerEntity": {
+    "offsets": {
+      "windows": 28,
```
b0ink, line 138: Unable to confirm this myself, but windows offset should be `27` according to [CS2Fixes' gamedata](https://github.com/Source2ZE/CS2Fixes/blob/e51d515719dfae562ff9595497fc44caff6bbfa2/gamedata/cs2fixes.games.txt#L403-L407)

schwarper: Corrected it, thanks

KillStr3aK (approved):

roflmuffin (approved):

PR #644, opened 2024-10-25
Updated CanAcquire signature after the 24/10/24 update

Dliix66:


KillStr3aK (approved):
😎

stefanx111:
After 28/10/24 update, the valid sig for linux is `55 48 89 E5 41 57 41 56 41 55 49 89 CD 41 54 53 48 83 EC`, the one that is already in gamedata.
So this PR is no longer needed.

Dliix66:
> After 28/10/24 update, the valid sig for linux is `55 48 89 E5 41 57 41 56 41 55 49 89 CD 41 54 53 48 83 EC`, the one that is already in gamedata. So this PR is no longer needed.

Correct 👍🏻

PR #661, opened 2024-11-06
CBaseEntity player teleport adjustment update

schwarper:


KillStr3aK:
its more like an adjustment/DX update than a fix

roflmuffin (approved):

PR #663, opened 2024-11-07
Added GetArgs

schwarper:
```csharp
[ConsoleCommand("css_testx")]
public void OnTest(CCSPlayerController player, CommandInfo info)
{
    var args = info.GetArgs();

    Console.WriteLine($"TESTX args count => {args.Count}");

    for (int i = 0; i < args.Count; i++)
    {
        Console.WriteLine($"{i}. {args[i]}");
    }
}

[ConsoleCommand("css_testy")]
public void OnTest2(CCSPlayerController player, CommandInfo info)
{
    var args = info.GetArgs(0, 3);

    Console.WriteLine($"TESTX args count => {args.Count}");

    for (int i = 0; i < args.Count; i++)
    {
        Console.WriteLine($"{i}. {args[i]}");
    }
}
```

image


schwarper:
https://github.com/roflmuffin/CounterStrikeSharp/pull/504

I improved this a bit.

KillStr3aK:
I'd also like the enumerator version maybe? from #504

schwarper:
> I'd also like the enumerator version maybe? from #504

I thought that getting arg once and split them is better than using GetArg multiple times. However, it might be better.

roflmuffin:
I think perhaps we are better off just having an `IEnumerable<string> Args =>` that yields in a for loop without any array access, and just pay the performance cost of multiple calls to `GetArg(i)`, that way the user can use regular methods to skip/take or do array accesses from the enumerable.

WidovV:
GetValue doesn't quite follow the naming convention since GetArg is a thing, hence GetArgs were the name for the method.
I do agree that calling GetArg multiple times instead of splitting ArgString seems like unnecessary performance waste, however I do not think that the first argument should be skipped. An example of usage could be a command listener for every chat message and only wanted to log messages starting with `!` or `/`

roflmuffin:
Have actually setermin

roflmuffin:
Have we actually determined that the performance of get arg would actually be worse than the string split? The con command on the native side would already know each arg so it's fairly cheap for it to make a copy of the arg and return it back to c#. I'd be interested in a benchmark comparing the two.

WidovV:
I've have contacted you over dm's on discord regarding how to setup a benchmark to test it 😄

schwarper:
```csharp
public static IEnumerable<string> GetArgs(CommandInfo info, int startIndex = 0, int endIndex = -1)
{
    if (string.IsNullOrEmpty(info.ArgString))
        return [];

    string[] args = info.ArgString.Split(' ', StringSplitOptions.RemoveEmptyEntries);
    int lastIndex = args.Length - 1;

    startIndex = Math.Clamp(startIndex, 0, lastIndex);
    endIndex = Math.Clamp(endIndex < 0 ? lastIndex : endIndex, startIndex, lastIndex);

    string[] selectedArgs = startIndex == endIndex
        ? [args[startIndex]]
        : args[startIndex..(endIndex + 1)];

    return selectedArgs.Select(arg =>
        arg.Length >= 2 && arg[0] == '"' && arg[^1] == '"'
            ? arg[1..^1]
            : arg
    );
}

public static IEnumerable<string> GetArgsYield(CommandInfo info)
{
    if (info.ArgCount == 1)
        yield break;

    for (int i = 1; i < info.ArgCount; i++)
        yield return info.GetArg(i);
}
```

test1
test2

schwarper:
image

PR #665, opened 2024-11-09
Added hitgroup to CTakeDamageInfo

schwarper:
image

```csharp
public static HookResult OnTakeDamage(DynamicHook hook)
{
    CEntityInstance entity = hook.GetParam<CEntityInstance>(0);

    if (entity.DesignerName != "player")
    {
        return HookResult.Continue;
    }

    var info = hook.GetParam<CTakeDamageInfo>(1);

    Server.PrintToChatAll($"HEY {info.Attacker.Value?.DesignerName} -> {info.GetHitGroup()}");
    return HookResult.Continue;
}
```

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Model/CTakeDamageInfo.cs
+++ b/managed/CounterStrikeSharp.API/Core/Model/CTakeDamageInfo.cs
@@ -1,0 +1,30 @@
+﻿namespace CounterStrikeSharp.API.Core;
+
+public partial class CTakeDamageInfo
+{
+    /// <summary>
+    /// Retrieves the hitgroup
+    /// </summary>
+    /// <returns>
+    /// Returns a <see cref="HitGroup_t"/> enumeration representing the player's current hit group,
+    /// or <see cref="HitGroup_t.HITGROUP_INVALID"/> if the hit group cannot be determined.
+    /// </returns>
+    public unsafe HitGroup_t HitGroup()
+    {
+        nint v4 = *(nint*)(this.Handle + 0x78);
+
+        if (v4 == nint.Zero)
+        {
+            return HitGroup_t.HITGROUP_INVALID;
+        }
+
+        nint v1 = *(nint*)(v4 + 16);
+
+        if (v1 == nint.Zero)
+        {
+            return HitGroup_t.HITGROUP_GENERIC;
+        }
+
+        return (HitGroup_t)(*(uint*)(v1 + 56));
+    }
+}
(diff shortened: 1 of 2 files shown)


roflmuffin:
The changed file is a generated file so you'll need to make a new file with a partial class to extend the generated code

managed/CounterStrikeSharp.API/Core/Schema/Classes/CTakeDamageInfo.g.cs
```diff
@@ -101,5 +101,4 @@ public CTakeDamageInfo (IntPtr pointer) : base(pointer) {}
 	// m_bInTakeDamageFlow
 	[SchemaMember("CTakeDamageInfo", "m_bInTakeDamageFlow")]
 	public ref bool InTakeDamageFlow => ref Schema.GetRef<bool>(this.Handle, "CTakeDamageInfo", "m_bInTakeDamageFlow");
-
-}
+}
```
roflmuffin, line 104: Can you please reset this file so its no longer in the diff?

managed/CounterStrikeSharp.API/Core/Model/CTakeDamageInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        nint v1 = *(nint*)(v4 + 16);
+
+        if (v1 == nint.Zero)
+        {
+            return HitGroup_t.HITGROUP_GENERIC;
+        }
+
+        return (HitGroup_t)(*(uint*)(v1 + 56));
+    }
```
roflmuffin, line 29: You should be able to convert this to safe code (untested):

```suggestion
    public HitGroup_t HitGroup()
    {
        IntPtr v4 = Marshal.ReadIntPtr(handle, 0x78);

        if (v4 == nint.Zero)
        {
            return HitGroup_t.HITGROUP_INVALID;
        }

        IntPtr v1 = Marshal.ReadIntPtr(v4, 16

        if (v1 == nint.Zero)
        {
            return HitGroup_t.HITGROUP_GENERIC;
        }

        return (HitGroup_t)Marshal.ReadInt32(v1, 56);
    }
```

schwarper: @roflmuffin yes it's working.

KillStr3aK:
this also could be a property instead? (by name)

or could be 'GetHitGroup'?

schwarper:
> this also could be a property instead? (by name)
> 
> or could be renamed* 'GetHitGroup'?

I guess GetHitGroup is better than HitGroup.

KillStr3aK:
> > this also could be a property instead? (by name)
> 
> > 
> 
> > or could be renamed* 'GetHitGroup'?
> 
> 
> 
> I guess GetHitGroup is better than HitGroup.

indeed, but 'HitGroyp' also stands good if its a property, not a method

roflmuffin (approved):

PR #695, opened 2024-11-28
feat: client message.

jaredliu2018:
Add client messages recv/send.
Server can hook any message(CLC_Messages) from client.
Also server can create a new client message and send it to client.
All complex logic is implemented in c++ side. Use it via simple API  like "HookClientMessage()" and "message.Send()".


configs/addons/counterstrikesharp/gamedata/gamedata.json
```diff
@@ -265,5 +265,11 @@
       "windows": 584,
       "linux": 584
     }
+  },
+  "CServerSideClient": {
+     "offsets": {
+       "windows": 8,
+       "linux": -64
+     }
```
KillStr3aK, line 273: #563 might need a revision

KillStr3aK:
Well this pr achieves way more stuff than described

can you regenerate the native methods with respecting the codestyle we use for them? (for e.x.: `UnhookClientmessage` -> `UnhookClientMessage`)

src/core/managers/clientmessage_manager.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    ClientMessageManager::~ClientMessageManager() {}
+
+    void ClientMessageManager::OnAllInitialized()
+    {
+        const char* name = "CServerSideClient";
+        void* serverSideClientVTable = reinterpret_cast<void*>(modules::engine->FindVirtualTable(name, globals::gameConfig->GetOffset(name)));
+        if (serverSideClientVTable == nullptr)
+        {
+            CSSHARP_CORE_ERROR("Failed to find signature for \'%s\'", name);
```
KillStr3aK, line 29: `Failed to find vtable for '%s'`

configs/addons/counterstrikesharp/gamedata/gamedata.json
jaredliu2018: Do I need change the game data? I think the offset of both os is correct and I have test it.

src/core/memory_module.cpp
```diff
@@ -562,4 +582,204 @@ void* CModule::FindSymbol(const std::string& name)
     CSSHARP_CORE_ERROR("Cannot find symbol {}", name);
     return nullptr;
 }
+
+Sections* CModule::GetSection(const std::string_view name)
```
Nukoooo, line 586: you can literally get everything you need in dl_iterate_phdr, or looping through PEB, copy-pasting these is making this code base even more bloated, same as SignatureIterator and the rest /shrug

jaredliu2018: Implement FindVirutalTable in another way, I think that is necessary.

Issue #699, opened 2024-12-02
[BUG] @domain/root flag does not work with RequiresPermissionsOr annotation


rothadamcz:
When admin has `@domain/root` flag, it is ignored when using RequiresPermissionsOr annotation.

# Example
- Admin has only 1 flag (`@css/root`)
- He is not able to execute css_ban (command implementation bellow), because he does not have `@css/ban` nor `@something/else` flag.

```cs
public override void Load(bool hotReload)
{
    AddCommand("css_ban", "Ban a player", OnCommand);
}

[RequiresPermissionsOr("@css/ban", "@something/else")]
[CommandHelper(minArgs: 1, usage: "[id]", whoCanExecute: CommandUsage.CLIENT_ONLY)]
public static void OnCommand(CCSPlayerController? caller, CommandInfo info)
{
    if (caller == null) return;

    info.ReplyToCommand("Hello World!");
}
```

Confirmed on API version: v291 (1.0.291+6349c11)

rothadamcz:
To clarify,
I think the intended behavior should be that access to the command should be granted, since required permission is `@css/ban` OR `@something/else` and since admin has `@css/root` which enables all `@css` flags including `@css/ban`, the requirement is met.

zonical:
Kia ora! I've been able to replicate the bug in the test suite, and I've found the cause of the issue in the code (which I wrote ages ago).

crashzk:
Any news about fix @zonical?

zonical:
> Any news about fix [@zonical](https://github.com/zonical)?

Hi! Sorry for the lack of work on this. I've been busy with work. I'll re-look into this in a few days.

PR #700, opened 2024-12-04
Fix handling of root flags in RequirePermissionsOr

zonical:
Addresses #699.


managed/CounterStrikeSharp.API/Modules/Admin/RequiresPermissionsOr.cs
```diff
@@ -11,31 +11,28 @@ public class RequiresPermissionsOr : BaseRequiresPermissions
     {
         public RequiresPermissionsOr(params string[] permissions) : base(permissions) { }
 
-        public override bool CanExecuteCommand(CCSPlayerController? caller)
+        public override bool CanExecuteCommand(SteamID? steamID)
         {
-            if (caller == null) return true;
-            if (AdminManager.PlayerHasCommandOverride(caller, Command))
+            if (steamID == null) return true;
```
KillStr3aK, line 16: is this reliable?

zonical: It should be. SteamID null checks are peppered throughout core AdminManager functions, including PlayerHasCommandOverride itself, and I haven't seen any bug reports these functions failing.

managed/CounterStrikeSharp.API/Modules/Admin/BaseRequiresPermissions.cs
```diff
@@ -27,13 +27,15 @@ public BaseRequiresPermissions(params string[] permissions)
             Command = "";
         }
 
-        public virtual bool CanExecuteCommand(CCSPlayerController? caller)
+        public virtual bool CanExecuteCommand(SteamID? steamID)
```
roflmuffin, line 30: Is this a public API? We might need to keep the old prototype around for backwards compatibility, and then proxy it to the new method using `caller.AuthorizedSteamID`

zonical: Addressed in cf025f0.

zonical:
Apologies for letting this branch go stale. Life has gotten in the way of me being able to complete this, and a new bug in AdminManager came to light from r3tr0attack, which I've included a fix and some new tests for in this PR.

roflmuffin (dismissed):

PR #720, opened 2024-12-24
Fix RemoveMapChangeTimers 

### LINKED ISSUE #704: Timers are improperly iterarated in RemoveMapChangeTimers
`RemoveMapChangeTimers` is iterating a vector while possibly removing from said vector, the current code does not shift the iterator correctly when an entry is removed and the iteration can skip entries. I would suggest to rewrite the entire function to use `std::erase_if` to kill all the active timers instead of manually iterating and calling KillTimer, this both fixes the issue and improves performance as you don't have to shift the entire vector every iteration.

https://github.com/roflmuffin/CounterStrikeSharp/blob/6349c11d07af8ba95d280e4a0f994ccb46053ef5/src/core/timer_system.cpp#L173C5-L179C6

Test plugin:
Run a dedicated server with the plugin loaded, load any map and then run `start_timer_test`

```cs
using CounterStrikeSharp.API;
using CounterStrikeSharp.API.Core;
using CounterStrikeSharp.API.Modules.Timers;

public class TestTimers : BasePlugin
{
    public override string ModuleName => "Timer Tests";
    public override string ModuleDescription => "";
    public override string ModuleAuthor => "Poggu";
    public override string ModuleVersion => "0.0.1";
    public bool MapChangeStarted = false;
    public bool MapChanged = false;

    public void TimerHandler(int timerNum)
    {
        Console.WriteLine($"Timer{timerNum} Tick");

        if (MapChanged)
        {
            Console.ForegroundColor = ConsoleColor.Red;
            Console.WriteLine("[!] Test Failed!");
            Console.ResetColor();
            MapChanged = false;
        }
    }
    
    {
        Console.WriteLine($"{ModuleName} loaded successfully!");
        
        AddCommand("start_timer_test", "", (player, info) =>
        {
            /* Test Preamble */
            Console.WriteLine("\n\n\n\n----------------------------------------------------------------------");
            Console.WriteLine("Starting 5 timers, disabling server hibernation and ALL LOG CHANNELS");
            Server.ExecuteCommand("log_verbosity Console off; log_verbosity - off; sv_hibernate_when_empty 0;");
            
            Console.WriteLine("Changing map in 10 seconds, no timers **should** run after map change.\n----------------------------------------------------------------------\n\n\n");
            AddTimer(15.0f, () =>
            {
                Console.WriteLine("\n\n---------\nChanging map\n---------\n\n\n");
                Server.ExecuteCommand("map de_dust2");
                MapChangeStarted = true;
            });
            
            /* Test Timers */
            AddTimer(3.0f, () =>
            {
                TimerHandler(1);
            }, TimerFlags.STOP_ON_MAPCHANGE | TimerFlags.REPEAT);
            AddTimer(3.0f, () =>
            {
                TimerHandler(2);
            }, TimerFlags.STOP_ON_MAPCHANGE | TimerFlags.REPEAT);
            AddTimer(3.0f, () =>
            {
                TimerHandler(3);
            }, TimerFlags.STOP_ON_MAPCHANGE | TimerFlags.REPEAT);
            AddTimer(3.0f, () =>
            {
                TimerHandler(4);
            }, TimerFlags.STOP_ON_MAPCHANGE | TimerFlags.REPEAT);
            AddTimer(3.0f, () =>
            {
                TimerHandler(5);
            }, TimerFlags.STOP_ON_MAPCHANGE | TimerFlags.REPEAT);
        });
        
        RegisterListener<Listeners.OnMapStart>(mapName =>
        {
            Console.WriteLine($"Map {mapName} loaded!");

            if (MapChangeStarted)
                MapChanged = true;
        });
    }
}
```

roflmuffin:
Closes #704

### DIFF
--- a/src/core/timer_system.cpp
+++ b/src/core/timer_system.cpp
@@ -30,14 +30,15 @@
  */
 
 #include "core/timer_system.h"
-#include <algorithm>
 
 #include <public/eiface.h>
 
+#include <algorithm>
+
 #include "core/globals.h"
 #include "core/log.h"
-#include "scripting/callback_manager.h"
 #include "core/managers/player_manager.h"
+#include "scripting/callback_manager.h"
 
 namespace counterstrikesharp {
 namespace timers {
@@ -46,8 +47,7 @@ double timer_next_think = 0.0f;
 } // namespace timers
 
 timers::Timer::Timer(float interval, float exec_time, CallbackT callback, int flags)
-    : m_interval(interval), m_exec_time(exec_time), m_flags(flags), m_kill_me(false),
-      m_in_exec(false)
+    : m_interval(interval), m_exec_time(exec_time), m_flags(flags), m_kill_me(false), m_in_exec(false)
 {
     m_callback = globals::callbackManager.CreateCallback("Timer");
     m_callback->AddListener(callback);
@@ -76,7 +76,8 @@ void TimerSystem::OnShutdown()
 
 void TimerSystem::OnLevelEnd()
 {
-    if (on_map_end_callback && on_map_end_callback->GetFunctionCount()) {
+    if (on_map_end_callback && on_map_end_callback->GetFunctionCount())
+    {
         on_map_end_callback->ScriptContext().Reset();
         on_map_end_callback->Execute();
     }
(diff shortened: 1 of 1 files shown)


Poggicek (approved):
Looks good, the new implementation doesn't respect m_in_exec like the old one, but I don't think it's needed in the first place, Killing a timer while executing it shouldn't ever happen if the timers aren't killed from a new thread, which they aren't.

PR #728, opened 2025-01-05
Make NetworkedVector support Vector and QAngle

Yarukon:
as the title said
only tested QAngle but Vector should work fine


KillStr3aK:
just mentioning #613

roflmuffin (approved):

PR #743, opened 2025-01-19
feat: handle quotes for `FakeConVar<string>`

ianlucas:
When trying to set URLs into `FakeConVar`s, the server doesn't parse it properly.

```sh
my_string_convar https://example.com
# my_string_convar.Value = "https:"

my_string_convar "https://example.com"
# my_string_convar.Value = "\"https://example.com\""
```

After this PR, the second case should work as expected. The first case seems to be the way the server parses the argument once we send the command, so nothing much we can do about it. The value needs to be wrapped into quotes to be extracted from them.

```sh
my_string_convar https://example.com
# my_string_convar.Value = "https:"

my_string_convar "https://example.com"
# my_string_convar.Value = "https://example.com"
```


M-archand:
Love it, pls merge rofl :)

roflmuffin (approved):

Issue #454, opened 2024-05-11
A method to control the characters limitation of a centerhtmlmenu item


samyycX:
Currently we have [CenterHtmlMenu.cs#L61](https://github.com/roflmuffin/CounterStrikeSharp/blob/02d5191e746aa08fb912746ceac61d4c891f9146/managed/CounterStrikeSharp.API/Modules/Menu/CenterHtmlMenu.cs#L61) and [CenterHtmlMenu.cs#72](https://github.com/roflmuffin/CounterStrikeSharp/blob/02d5191e746aa08fb912746ceac61d4c891f9146/managed/CounterStrikeSharp.API/Modules/Menu/CenterHtmlMenu.cs#L72) to limit the character amount. 

I think it is because the original developer want to ensure items are displayed in one line. (https://github.com/roflmuffin/CounterStrikeSharp/pull/268#issuecomment-1901229101)
But with such limitation, the html labels in text will be counted too, even if they don't change the length of the text,
and using other characters (like Chinese characters) will cause the view to be malformed too.

It would be great if we can have a method to control these limitations, and please tell me if there are more reasons to have these limitations.

PR #747, opened 2025-01-22
Remove the characters limit for CenterHtmlMenu

samyycX:
The reason is in #454


roflmuffin:
Thanks for the contribution, I think it does make sense to remove the character count and let plugin authors decide for themselves how it appears.

Would you be able to revert some of the formatting changes in this PR?

samyycX:
> Thanks for the contribution, I think it does make sense to remove the character count and let plugin authors decide for themselves how it appears.
> 
> Would you be able to revert some of the formatting changes in this PR?

fixed, i believe there is the only format problem here

roflmuffin (approved):

PR #750, opened 2025-01-27
fix: errors when reloading generated configuration

qstage:
The comment at the beginning of the automatically generated configuration results in an error when reloading.
https://github.com/roflmuffin/CounterStrikeSharp/blob/main/managed/CounterStrikeSharp.API/Modules/Extensions/PluginConfigExtensions.cs#L78


KillStr3aK:
I believe you should drop the first two commits as they are unrelated to the goal

PR #757, opened 2025-02-04
feat: add `ListenerHandlerAttribute<T>`

qstage:
Adds a feature listener registration via attributes. Loading & unloading of the plugin has been verified. If `T` represents a delegate that does not have a `ListenerName` attribute, will be thrown `ArgumentException`, similar to how `RegisterListener` works.

```c#
public class Plugin : BasePlugin
{
    public override string ModuleName => "TestPlugin";

    [ListenerHandler<Listeners.OnServerPreWorldUpdate>]
    public void OnServerPreWorldUpdate(bool simulating)
    {
        if (Server.TickCount % 128 == 0)
        {
            Logger.LogInformation("Tick: {count}", Server.TickCount);
        }
    }

    [ListenerHandler<Listeners.OnClientPutInServer>]
    public void OnClientPutInServer(int playerSlot)
    {
        var player = Utilities.GetPlayerFromSlot(playerSlot);

        if (player == null || player.IsBot) return;
        
        player.PrintToChat("Welcome to the server!");
    }
}
```

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Attributes/Registration/ListenerHandlerAttribute.cs
+++ b/managed/CounterStrikeSharp.API/Core/Attributes/Registration/ListenerHandlerAttribute.cs
@@ -1,0 +1,9 @@
+using System;
+
+namespace CounterStrikeSharp.API.Core.Attributes.Registration;
+
+[AttributeUsage(AttributeTargets.Method)]
+public class ListenerHandlerAttribute<T> : Attribute
+    where T: Delegate
+{
+}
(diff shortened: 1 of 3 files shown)


roflmuffin:
> I'm thinking about naming the attribute just 'Listener' rather than 'ListenerHandler', not sure about @roflmuffin

I could go either way, because of the cursed nature of the Listeners parent class name, it might look terrible just having `[Listener<Listeners.OnClientPutInServer>]`.

roflmuffin (approved):

PR #758, opened 2025-02-04
Restore HTML menu max length, but as an optional feature and improved the trimming with HTML banners

Dliix66:
This PR is following #747

It adds the ability to enable or disable the limit. But it also improves the trimming by not taking HTML banners into account, but it also closes the open banners after trimming to avoid any visual issues.

Example: 
 "<b>Hello</b> <font color='green'>World</font>! This is a test." trimmed to 8 characters will be "<b>Hello</b> <font color='green'>Wo</font>"

To do this, add this in your config:
"MaxHtmlMenuTitleLength": 0, // 0 = unlimited, previous value = 32
"MaxHtmlMenuOptionLength": 0, // 0 = unlimited, previous value = 26`

### DIFF
--- a/configs/addons/counterstrikesharp/configs/core.example.json
+++ b/configs/addons/counterstrikesharp/configs/core.example.json
@@ -7,6 +7,6 @@
     "ServerLanguage": "en",
     "UnlockConCommands": true,
     "UnlockConVars": true,
-    "MaxHtmlMenuTitleLength": 0, // 0 = unlimited, previous value = 32
-    "MaxHtmlMenuOptionLength": 0, // 0 = unlimited, previous value = 26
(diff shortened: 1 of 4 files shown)


roflmuffin:
Not sure this should really be a core config, given that the plugin author themselves would not be able to predict the behaviour of their menu based on this setting. If this is a desired functionality of the menu, I think it should be opt-in at the plugin level rather than it coming from CSS

ValMadBox:
Also, my trimming improvement example was not displayed correctly cause I forgot to wrap it 😓 
Here it is:

`"<b>Hello</b> <font color='green'>World!</font> This is a test." trimmed to 8 characters will be "<b>Hello</b> <font color='green'>Wo</font>"`

PR #772, opened 2025-02-09
feat: added worldtext_menu. (ScreenMenu)

T3Marius:
I've added ScreenMenu to MenuManager.

It works same as CenterHtmlMenu and has the necessary checks and clears thanks to Poggu.

I don't know if it's a problem so i've made it confiugrable but i've added the Create function in CreateScreenMenu.cs where people can create hud text if they set the isMenu to false and when isMenu is true it adds a backround to the text.

Only things that are need to be done in my opinion are using the spectating player viewmodel to acces the menu while spectating players. For those in spectator team or free cam i don't have any idea.

[W/S] to scroll and [E] to select.


Dliix66:
Can you remove your formatting? This makes the PR hard to read 😅

roflmuffin:
My gut feeling is that this implementation is a little too involved to go into the core of CS# and might be better off as a library published on nuget

T3Marius:
> My gut feeling is that this implementation is a little too involved to go into the core of CS# and might be better off as a library published on nuget

Got it

PR #804, opened 2025-03-17
feat: add loading of toml support

### LINKED ISSUE #773: Support TOML as a First-Class Configuration Format
### Discussed in https://github.com/roflmuffin/CounterStrikeSharp/discussions/764

<div type='discussions-op-text'>

<sup>Originally posted by **Mesharsky** February  6, 2025</sup>
TOML is becoming a popular choice for configuration files due to its simplicity and readability. JSON is machine-friendly but can be tedious for humans, especially when editing configs manually. Adding support for TOML would improve usability without removing JSON compatibility.  

## Why TOML?  
- **More Readable & Human-Friendly** – TOML is designed for configs, with a cleaner and more intuitive syntax than JSON.  
- **Supports Comments** – Unlike JSON, TOML allows comments (`# like this`), making it easier to document configurations. (I know JSON also supports comments, but they are shit)  
- **Less Error-Prone** – No strict formatting like JSON (e.g., missing commas won't break everything).  
- **Better for Complex Configurations** – Handles nested structures, lists, and tables more naturally.  

## Implementation Proposal  
1. **Support both JSON & TOML** – Detect format automatically and load the corresponding config.  
2. **Default to TOML, keep JSON as fallback** – New setups use TOML, but JSON remains supported.  
3. **Full migration to TOML** – Deprecate JSON in favor of TOML (only if widely preferred).  

Myself and also people that are using, for example, my plugins find TOML much more friendly and easier to deal with. In my opinion, JSON was a bad choice from the start.  
</div>

roflmuffin:
Adds support for basic TOML config parsing for plugin configs.

- If a `PluginName.toml` file is found, it will read that using `Tomlyn`.
- If a `PluginName.example.toml` file is found, it will copy the example to `PluginName.toml` and then read it.
- Otherwise, fall back to existing JSON behaviour (read .json, copy example, same as above).
- If no existing example or real config file are found, it will default to saving a `.json` file with the defaults to maintain similar behaviour as before.
- There is currently no way to provide TOML options to the config serializer/deserializer, so it will use our hard coded defaults which preserves names for property names.

Closes #773


Mesharsky:
@roflmuffin 

I would add support for Preserve PascalCase

```
var tomlContent = File.ReadAllText(configPath);

var tomlModelOptions = new TomlModelOptions
{
    ConvertPropertyName = name => name // Preserve PascalCase (disable auto conversion)
};

Config = Toml.ToModel<SpeedHackConfig>(tomlContent, options: tomlModelOptions);
```

Other than that, finally <3

roflmuffin:
@Mesharsky, I think this is good to go and I've tested it locally and everything appears to work correctly. Would you like to pull down the artifacts and test for yourself before merge or all good?

Mesharsky:
@roflmuffin Works perfectly! Nice implementation

Mesharsky:
@roflmuffin Oh, i thought it's working, is my implementation wrong?
Inside:

/game/csgo/addons/counterstrikesharp/configs/plugins/AWPManager 

I have AWPManager.toml
Despite that it's still generating JSON file, not reading values from TOML file, so i guess fallback mechanism is in place

Config class is as follows

using CounterStrikeSharp.API.Core;

```
namespace AWPManager
{
    public class AWPManagerConfig : BasePluginConfig
    {
        public DatabaseConfig Database { get; set; } = new();
        public CoreConfig Core { get; set; } = new();
        public AnimationConfig Animation { get; set; } = new();
        public Dictionary<string, GameModeConfig> GameModes { get; set; } = new();
        public StatisticsConfig Statistics { get; set; } = new();
        public WeaponConfig Weapons { get; set; } = new();
    }
```

Plugin.cs just registers it globally as:

`public AWPManagerConfig Config { get; set; } = new();`

Debugs i have is basically this:

```
[AWPManager] Configuration loaded and validated successfully.
[AWPManager] ERROR: Database connection test failed - Unable to connect to any of the specified MySQL hosts.
[AWPManager] ERROR: Could not establish database connection
[AWPManager] Registered game mode: Standard AWP (standard)
[AWPManager] Registered game mode: No-Scope Challenge (noscope)
[AWPManager] Registered 2 game modes
[AWPManager] StatisticsManager initialized
```

And it creates .json file every time, called: AWPManager.json which is wrong?

roflmuffin:
That location looks right, it shouldn't try to create a json if the toml exists, are you definitely on 313+?

Mesharsky:
@roflmuffin You are right, for some reason the package did not automatically update for me, what a headache, my bad.
Works as it should

PR #808, opened 2025-03-18
feat: Implement `EmitSoundFilter`

samyycX:
The `EmitSound` feature has been a frequently mentioned problem in discord for a while. Due to the params of sdk structs, it's hard to implement it in a pure C# plugin. This PR implement it in the native side and provide a easy-to-use interface in managed side. Tested on windows, works fine without any problem.
Part of the code and signature is from cs2fixes.

Btw im not quite sure how to design the c# side api, if you have any better idea, please tell me.

## Add
- `uint CBaseEntity.EmitSound(string soundName, float volume = 1f, float pitch = 0f)`
- `uint CBaseEntity.EmitSoundWithFilter(RecipientFilter, string soundName, float volume = 1f, float pitch = 0f)`

## Usage
```csharp
// To all player
entity.EmitSound("CT_Water.StepLeft", volume: 10f, pitch: 0f);

// To specific player
RecipientFilter filter = [ player ];
player.EmitSound("CT_Water.StepLeft", filter, 1f);
```

PS: the pitch is a int type param from the struct, but cs2fixes made it a float, not sure if theres a problem but other feature works fine.


src/core/managers/entity_manager.cpp
```diff
@@ -265,4 +274,14 @@ void DetourFireOutputInternal(CEntityIOOutput* const pThis, CEntityInstance* pAc
...
 }
 
+SndOpEventGuid_t EntityEmitSoundFilter(IRecipientFilter& filter, uint32 ent, const char* pszSound, float flVolume, float flPitch)
+{
+    EmitSound_t params;
+    params.m_pSoundName = pszSound;
+    params.m_flVolume = flVolume;
+    params.m_nPitch = flPitch;
+
+    return CBaseEntity_EmitSoundFilter(filter, ent, params);
```
roflmuffin, line 284: Since we may fail to find the signature when we initialise the entity manager, it may be worth adding a check in here that will only call if it it is a valid pointer, that way the worst case is that sounds don't play vs. a possible crash when there is no valid signature on game update.

samyycX: Thanks for pointing out, should be solved now

samyycX:
It seems that the volume is not working for some unknown reason, we may need to implement `EmitSoundParam` to solve it. This has to be resolved before we merge it.

samyycX:
> It seems that the volume is not working for some unknown reason, we may need to implement `EmitSoundParam` to solve it. This problem has to be reviewed before we merge it.

@roflmuffin This is now confirmed as a issue from valve's code and we can utilize a C# plugin to resolve it. There's should be no problem about these core codes.

roflmuffin (approved):

jxn78:
[09:22:24.537] CSSharp: Failed to find signature for 'CBaseEntity_EmitSoundFilter'

roflmuffin:
> [09:22:24.537] CSSharp: Failed to find signature for 'CBaseEntity_EmitSoundFilter'
> 
> Linux
> 
> @roflmuffin

Fixed in v310

PR #826, opened 2025-03-27
Update player_chat args

Yarukon:
should correspond to the latest changes for the gameevents.
don't know what playerid should be, so i'll keep using the player slot.


roflmuffin:
I forgot we are firing this manually, maybe we should exempt this event from codegen and manually wire it up how we had it before?

Yarukon:
closed since revert

Issue #833, opened 2025-04-01
Update CCSPlayer_WeaponServices_CanUse signature


darkerz7:
```
  "CCSPlayer_WeaponServices_CanUse": {
    "signatures": {
      "library": "server",
      "windows": "48 89 5C 24 ? 48 89 6C 24 ? 48 89 74 24 ? 57 48 83 EC ? 48 8B 01 48 8B F2",
      "linux": "55 48 8D 15 ? ? ? ? 48 89 E5 41 55 49 89 F5 41 54 49 89 FC 53 48 83 EC ? 48 8B 07 48 8B 80 ? ? ? ? 48 39 D0 0F 85 ? ? ? ? 80 BF"
    }
  },
```

from -> https://github.com/Source2ZE/CS2Fixes/commit/e1265b54cef49dd00d4476dd57609e8b01ae7216

PR #837, opened 2025-04-02
Update CCSPlayer_WeaponServices_CanUse signature

imi-tat0r:


KillStr3aK:
are you sure that signature is right? it seems to be a relative address instead (I did not check, just making sure)

KillStr3aK (approved):
I did not run  any tests with the signature but the new one LGTM

darkerz7:
in cs2fixes there is a different signature https://github.com/Source2ZE/CS2Fixes/commit/ff4f1882c0317cb37282e4fa4d6431ca4047680c
also the signature for CS# is contained in the following issue https://github.com/roflmuffin/CounterStrikeSharp/issues/833

PR #873, opened 2025-05-21
Added validation in GetPlayers & added 2 new functions in WeaponServices

Cruze03:
**feat**: Additional validation in `GetPlayers()` .
**feat:** `DropWeapon` & `SelectWeapon` in WeaponServices.
**update:** `RemoveItemByDesignerName` with new DropWeapon function.

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Model/CBaseEntity.cs
+++ b/managed/CounterStrikeSharp.API/Core/Model/CBaseEntity.cs
@@ -25,6 +25,28 @@ public partial class CBaseEntity
             _angles, _velocity);
     }
 
+    /// <summary>
+    /// Check if entity is controller
+    /// </summary>
+    /// <exception cref="InvalidOperationException">Entity is not valid</exception>
+    public bool IsController()
+    {
+        Guard.IsValidEntity(this);
+
+        return VirtualFunction.CreateVoid<IntPtr>(Handle, GameData.GetOffset("IsPlayerController"))(Handle);
+    }
+
+    /// <summary>
+    /// Check if entity is pawn
+    /// </summary>
+    /// <exception cref="InvalidOperationException">Entity is not valid</exception>
+    public bool IsPawn()
+    {
+        Guard.IsValidEntity(this);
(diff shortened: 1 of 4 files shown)


managed/CounterStrikeSharp.API/Core/Model/CBaseEntity.cs
```diff
@@ -25,6 +25,28 @@ public void Teleport(Vector? position = null, QAngle? angles = null, Vector? vel
             _angles, _velocity);
     }
 
+    /// <summary>
+    /// Check if entity is controller
+    /// </summary>
+    /// <exception cref="InvalidOperationException">Entity is not valid</exception>
+    public bool IsController()
```
roflmuffin, line 32: Not sure there is huge value in having to maintain the offsets for these methods when there are other ways to do a pawn/controller check (namely a classname comparison).

Cruze03: True but I think ideal way to controller check would be this as I saw in cs2fixes, this was being used

managed/CounterStrikeSharp.API/Utilities.cs
```diff
@@ -97,12 +97,16 @@ public static bool RemoveItemByDesignerName(this CCSPlayerController player, str
         public static bool RemoveItemByDesignerName(this CCSPlayerController player, string designerName, bool shouldRemoveEntity)
         {
             CHandle<CBasePlayerWeapon>? item = null;
-            if (player.PlayerPawn.Value == null || player.PlayerPawn.Value.WeaponServices == null) return false;
+
+            if (player.PlayerPawn.Value == null
+            || !player.PlayerPawn.Value.IsValid
```
roflmuffin, line 102: This valid check is unnecessary, `.Value` will return null under the hood if `IsValid` is false anyway, so the second check is redundant.

Cruze03: oh nice, I didn't know that. I removed that check

PR #889, opened 2025-06-08
feat: Implement `GetRegister`

ELDment:
The `GetRegister` function may be redundant, but it is particularly important when making hooks for special functions. Due to the hook callback control flow, it is difficult to get the real value of volatile registers using pure C#. This PR implements this function in the Native layer and provides an easy-to-use interface on the managed side. It has been tested on Windows and runs perfectly without any problems.

Btw. I really hope it can be merged as it would greatly help our plugins development!‌ 👍💡

### Add
T DynamicHook.GetRegister(DHookRegister index)

### Usage
```CSharp
private MemoryFunctionVoid<nint, nint> unkfunc = new("48 89 5C 24 ? 57 48 81 EC ? ? ? ? 0F 29 74 24 ? 48 8B DA 0F 29 7C 24 ? 48 8D 54 24 ? 48 8B F9");

unkfunc.Hook((h) => {
	M128A m128A = h.GetRegister<M128A>(DHookRegister.XMM0);

	Console.WriteLine($"{h.GetRegister<nint>(DHookRegister.ECX_RCX)} - {h.GetParam<nint>(0)}"); // Same results
	Console.WriteLine($"Return：{m128A.GetDouble(0)}");

	return HookResult.Continue;
}, HookMode.Post);
```


managed/CounterStrikeSharp.API/Core/ScriptContext.cs
```diff
@@ -466,6 +466,21 @@ internal unsafe object GetResult(Type type, byte* ptr)
...
+
+				if (dataPointer == IntPtr.Zero)
+				{
+					// This means C++ returned a null pointer. Handle as appropriate.
+					// For now, throwing, as the original NRE was due to the generic path returning null for M128A.
+					throw new NativeException("Native function returned a null pointer for M128A result.");
+				}
+				// Marshal the data from the location pointed to by dataPointer
+				return Marshal.PtrToStructure<CounterStrikeSharp.API.Modules.Memory.M128A>(dataPointer);
+			}
```
KillStr3aK, line 482: I think you shouldn't use the full import path for the `M128A` type

```csharp
using CounterStrikeSharp.API.Modules.Memory;
```

```csharp
typeof(M128A)
```

```csharp
return Marshal.PtrToStructure<M128A>(dataPointer);
```

or if you'd like to make it clear where the type is coming from maybe use `Memory.M128A`

KillStr3aK:
LGTM, [CompatibilitySuppressions.xml](https://github.com/roflmuffin/CounterStrikeSharp/pull/889/files#diff-c004310e06f4f06b847fcef2a189c1d18cc0e9eae3f7d168b328aac388ddab34) should be removed tho

also the clang format should be fixed

ELDment:
~

ELDment:
~

PR #895, opened 2025-06-15
Improve FunctionReference trace logging with real user stack origin

SlynxCZ:
### Summary

This PR adds structured callback trace registration to CounterStrikeSharp by capturing the *actual user-level stack frame* that registered a delegate. 

This feature by itself does **not change runtime behavior** or add logging — it’s purely structural and performance-neutral. It prepares the internal architecture for advanced debugging and crash diagnostics without introducing risk to existing systems.

---

#### Old output (completely unhelpful):
`[23:40:30.477] CSSharp: Native 5df2e20d was invoked on a non-main thread`

This gives no information about which plugin or part of the code triggered it.

#### New trace (fully traceable via my external MM plugin):
`[CALLBACK TRACE] name='MyPlugin+<>c__DisplayClass7_0.<OnMapStart>b__0'
callerStack(top)='MyPlugin.OnMapStart @ MyPlugin.cs:42'`

This reveals exactly which delegate was registered and where — essential for diagnosing crashes, thread violations, or slow callbacks.

---

### Implementation

- Adds a filtered `StackTrace(2, true)` inside `FunctionReference.CreateWrappedCallback()`
- Skips internal frames (e.g. `CounterStrikeSharp`, `SafeExecutor`, compiler-generated closure types)
- Passes the first meaningful frame to `RegisterCallbackTrace(...)`
- Does **not** log anything directly — avoids console spam or runtime overhead

This is a backend-only improvement that enables external tooling support.

---

### AcceleratorCSS Integration

This structural change enables native integration with **AcceleratorCSS**, a Metamod plugin I’ve developed, which hooks into `RegisterCallbackTrace()`.

Once this PR is merged:
- `AcceleratorCSS` will be **fully compatible with all Linux builds** of CSS
- It will log the last N registered callbacks and write them to disk on crash (along with console history and map info)
- It dramatically improves post-mortem debugging, especially for SIGSEGV or thread-related crashes
- **Windows support is also planned** in future updates

The plugin is finished and will be released immediately after this PR is merged.

---

Let me know if you prefer the trace logic behind a config flag, or extracted into a utility. Otherwise, this is backward-safe and ready to merge.


SlynxCZ:
I actaully forgot to remove the `configs/addons/counterstrikesharp/configs/core.json` before creating pull request, this is not required in my functionaly, so i can change it later, also i forgot to say that i added some sigs & offsets that weren't there before, they are: 

- GetHammerUniqueId
- ChangeSubclass
- CAttributeList_SetOrAddAttributeValueByName
- CBaseModelEntity_SetBodygroup

src/scripting/callback_manager.cpp
```diff
@@ -36,22 +63,61 @@ void ScriptCallback::AddListener(CallbackT fnPluginFunction) { m_functions.push_
...
+    catch (...)
+    {
+        CSSHARP_CORE_WARN("Context is invalid (exception during access)");
+        return false;
+    }
 }
 
 void ScriptCallback::Execute(bool bResetContext)
 {
+    if (m_functions.empty())
```
roflmuffin, line 88: I don't think this is valid. We create global listeners which could execute with zero listeners.

src/core/managers/event_manager.cpp
```diff
@@ -261,17 +289,24 @@ bool EventManager::OnFireEventPost(IGameEvent* pEvent, bool bDontBroadcast)
...
+            EventOverride Override = { bDontBroadcast };
             pCallback->Reset();
             pCallback->ScriptContext().Push(pEventCopy);
-            pCallback->ScriptContext().Push(&override);
+            pCallback->ScriptContext().Push(&Override);
             pCallback->Execute();
 
-            globals::gameEventManager->FreeEvent(pEventCopy);
-            m_EventCopies.pop();
+            if (pEventCopy)
```
roflmuffin, line 301: While its good to include this event manager change in the PR, theres too many formatting changes in this file.

configs/addons/counterstrikesharp/gamedata/gamedata.json
```diff
@@ -1,4 +1,31 @@
 {
+  "GetHammerUniqueId": {
```
roflmuffin, line 2: If these offsets/sigs are not used in this PR then they should be raised separately. I'm hesitant to add anything that is not used by core as its just extra sigs we have to maintain on an update.

SlynxCZ:
All requested changes have been applied:

- Reverted unrelated formatting in event_manager.cpp
- Cleaned up unused sigs in gamedata.json
- Fixed Execute() logic to allow global listeners
- Ensured FreeEvent() is only called when the pointer is valid

Let me know if you’d prefer anything split into a separate PR or adjusted further.
Thanks again for the feedback!

roflmuffin:
LGTM, might need to fix the clang-format issues

SlynxCZ:
> LGTM, might need to fix the clang-format issues

Fixed the clang-format complaint from CI. Good to go now!

src/scripting/callback_manager.cpp
```diff
@@ -36,22 +63,55 @@ void ScriptCallback::AddListener(CallbackT fnPluginFunction) { m_functions.push_
...
+        CSSHARP_CORE_WARN("Context is invalid (exception during access)");
+        return false;
+    }
 }
 
 void ScriptCallback::Execute(bool bResetContext)
 {
+    if (!IsContextSafe())
+    {
+        ScriptContext().ThrowNativeError("ScriptCallback::Execute aborted due to invalid context");
```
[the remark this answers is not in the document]
SlynxCZ: As i remember, sometimes in newer css versions when was warmup and player which was alive changed team, where plugin with eventplayerspawn opened instantly any menu, it stopped recognizing inputs from OnKeyPress excluding the next, previous and close actions. This fixed it for me.

PR #905, opened 2025-06-27
Implement CVariant functions

zonical:
Implements functions related to `variant_t` / `CVariant` classes. New PR as the old branch went extremely stale.


roflmuffin:
Are the GameEvents file changes required?

zonical:
> Are the GameEvents file changes required?

Nope! Those should be reverted now :)
A thing worth noting is that `fieldtype_t` was updated to have fields such as `FIELD_INTEGER` and `FIELD_FLOAT` converted to `FIELD_INT32` and `FIELD_FLOAT32` respectively. This shouldn't have a *major* impact on things, as I believe the `FIELD_STTRING` type will cover most use cases. It's possible to add 64-bit-specific functions later on if required.

zonical:
@roflmuffin any progress on the review?

roflmuffin (approved):

PR #911, opened 2025-07-02
Fixes for game update 1.40.8.5 (03.07.2025)

AspectUnk:
The old `Host_Say` signature now points to a different function
image


Fess23:
Worked

KillStr3aK:
based on the discord messages we also need a new schema dump as the heavy armor has been removed from the game (not sure about what else) but this also gonna break plugins that accessed these schema members

AspectUnk:
@KillStr3aK It can be replaced with a dummy value to preserve backward compatibility, and also marked as obsolete. In the end, it's up to the community or the author to decide.

PR #922, opened 2025-07-07
Support operating `bytes` type protobuf field

samyycX:
As the title said, this PR allow plugins to read/write bytes type protobuf field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## What's Changed in v1.0.325
+* fix: revert internal stack trace for performance in [#927](https://github.com/roflmuffin/CounterStrikeSharp/pull/927) ([260015d](https://github.com/roflmuffin/CounterStrikeSharp/commit/260015d765f7feea8eb75773cc307dfe21c3a084))
+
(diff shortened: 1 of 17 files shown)


src/scripting/natives/natives_usermessages.cpp
```diff
@@ -238,6 +238,52 @@ static void PbReadString(ScriptContext& scriptContext)
     scriptContext.SetResult(returnValue.c_str());
 }
 
+static void PbReadBytes(ScriptContext& scriptContext)
```
roflmuffin, line 241: We should probably add another native to get the length of the protobuf field so that the C# side can automatically allocate the correct buffer length to prevent issues where 2048 bytes is too small and the user can't change this amount.

roflmuffin:
Have you got an example plugin/proto you are using to test that this is working?

samyycX:
> Have you got an example plugin/proto you are using to test that this is working?

Hook Usermessage 208 (CMsgSosStartSoundEvent) and read the `packed_params` field, it will be triggered when player make sounds

roflmuffin:
LGTM, I'll test it quickly later today

roflmuffin (approved):

PR #924, opened 2025-07-08
Remove internal stack traces and add native signature bindings

SlynxCZ:
### Summary

This PR removes the runtime stack trace collection from `RegisterCallbackTrace(...)` due to significant overhead, and instead adds native-level signature bindings with clean wrappers.

These changes are **runtime-neutral**, fully backward compatible, and improve performance, stability, and modding capability.

---

#### Removed:

`StackTrace(2, true)` was previously used to capture the origin of each registered delegate. However, testing revealed that it incurs a **high runtime cost** when many callbacks are registered (especially during map start or event hooks).

Since [`AcceleratorCSS`](https://github.com/SlynxCZ/AcceleratorCSS) already provides external callback tracing at runtime with near-zero overhead, this internal mechanism was redundant.

#### New Focus:

Instead of internal diagnostics, this PR focuses on **exposing native engine features** for plugin developers via clean C# wrappers.

---

### Added

New native functions were added to `VirtualFunctions` via signature binding:

- `CBaseModelEntity::SetBodygroup`
- `CAttributeList::SetOrAddAttributeValueByName`
- `CBasePlayerWeapon::ChangeSubclass`

#### Example usage:

```csharp
// Change bodygroup (e.g. show/hide gloves)
pawn.SetBodygroup("default_gloves", 1);

// Modify item attributes (StatTrak, paintkit, wear, seed, etc.)
item.AttributeList.SetAttribute("kill eater", 1337);
item.NetworkedDynamicAttributes.SetAttribute("paintkit", 180);

// Change weapon subclass to different defindex (e.g. switch M4A1 → M4A4)
weapon.ChangeSubclass(5027);
```

These are wrapped in:
- `CBaseModelEntity.SetBodygroup(...)`
- `CAttributeList.SetAttribute(...)`
- `CBasePlayerWeapon.ChangeSubclass(...)`

---

### AcceleratorCSS Compatibility

This PR keeps the system **fully compatible with [AcceleratorCSS](https://github.com/SlynxCZ/AcceleratorCSS)** — a native Metamod plugin for CSS that:

- Tracks all runtime C# callback invocations
- Logs the last N callbacks before a crash
- Writes full crash context to disk
- Supports `SIGSEGV`, native callstacks, and callback profiles

It now fully relies on the lightweight external trace mechanism, without duplicating effort inside CounterStrikeSharp itself.

Future versions of AcceleratorCSS will expand with:
- Windows crash support

---

### Implementation

- Removed expensive internal stack trace logic (`StackTrace(2, true)`)
- Added native signature bindings:
  - `SetBodygroup`
  - `SetOrAddAttributeValueByName`
  - `ChangeSubclass`
- Added safe wrappers to appropriate entity types:
  - `CBaseModelEntity.cs`
  - `CAttributeList.cs`
  - `CBasePlayerWeapon.cs`
- Tested and production-ready (used in my live FUNPLAY servers)

---

### Notes

- Does **not** log or change runtime behavior
- No config flags needed — safe by default
- Great base for modding (skins, agents, kill counter logic, model visibility)

Let me know if you prefer config gating or capability restriction. Otherwise, this is stable, validated, and good to merge.


roflmuffin:
It would be good to implement the changes to the stacktrace tracking, but there are too many other miscellaneous changes in this PR (including formatting).

Would it be possible to create a separate PR with only the script context changes and nothing else, and without the formatting changes.

roflmuffin:
I've reverted the internal stacktrace in #927

PR #925, opened 2025-07-09
feat: Add Automatic Gamedata Updates

roflmuffin:
Adds a basic automatic gamedata update system which will check the latest gamedata.json file from a CDN and download the new gamedata file if the ETag is different.


KillStr3aK:
don't forget to update the [docs page](https://docs.cssharp.dev/docs/reference/core-configuration.html) with the new core config entries

KillStr3aK (approved):

PR #930, opened 2025-07-10
fix: Remove requirement for https for game config updates

roflmuffin:
The CDN has been updated to not force https redirect, and we no longer link against libssl


lukepolo:
Curious , was it just to not force extra lib ?

PR #935, opened 2025-07-14
feat: add Convar<T> for real convar creation/deletion/access

roflmuffin:


managed/CounterStrikeSharp.API/Core/API.cs
```diff
@@ -218,6 +218,146 @@ public static void ReplicateConvar(int clientslot, string convarname, string con
 			}
 		}
 
+        public static void SetConvarFlags(ushort convar, ulong flags){
+			lock (ScriptContext.GlobalScriptContext.Lock) {
+			ScriptContext.GlobalScriptContext.Reset();
+			ScriptContext.GlobalScriptContext.Push(convar);
+			ScriptContext.GlobalScriptContext.Push(flags);
+			ScriptContext.GlobalScriptContext.SetIdentifier(0xB2BDCCBF);
```
ubbemir, line 226: Magic numbers should not be written by themselves. Maybe make them a constant with a name to clarify?

ubbemir (commented):
This looks awesome. Would be great to also have a list of known ConVars that utilize this to reduce mistake of accidentally using the wrong type with a ConVar.

KillStr3aK:
> This looks awesome. Would be great to also have a list of known ConVars that utilize this to reduce mistake of accidentally using the wrong type with a ConVar. This however would be fitting in another PR I suppose.

https://cs2.poggu.me/dumped-data/convar-list

managed/CounterStrikeSharp.API/Core/API.cs
KillStr3aK: these are hash calculated native identifiers

ubbemir:
> > This looks awesome. Would be great to also have a list of known ConVars that utilize this to reduce mistake of accidentally using the wrong type with a ConVar. This however would be fitting in another PR I suppose.
> 
> 
> 
> https://cs2.poggu.me/dumped-data/convar-list

Yes I know of that. But implement some of these at least into C# as statically typed to make use of the strength of static typing and static code analysis to reduce mistakes and catch erronous usage at compile time.

For example would it be suitable to do something similar to this?
```
public static class KnownConVars
{
    public static ConVar<bool> SvCheats = new ConVar<bool>("sv_cheats");
    public static ConVar<int> MpWarmupTime = new ConVar<int>("mp_warmuptime");
    public static ConVar<float> MaxSpeed = new ConVar<float>("sv_maxspeed");

    // Example: Add other ConVars here...
}
```

PR #943, opened 2025-07-17
feat: add Vector3 and QAngle explicit casts to System.Numerics.Vector3

roflmuffin:
Since `Teleport` is probably the biggest source of Vector based memory leaks, this adds a way to call the Teleport virtual method using `System.Numerics.Vector3` while still maintaining backwards compatibility.

To increase the ergonomics of modifying existing vectors coming out of the schema system (like with AbsOrigin), I've added explicit cast operators to allow you to easily convert to a numerics Vector3. This means the following code can be substituted to remove memory leak:

Before:
```cs
// Leaks because 2021 Roflmuffin sucks and `new Vector` allocates memory in native side and never frees it
pawn.Teleport(new Vector(pawn.AbsOrigin.X, pawn.AbsOrigin.Y, pawn.AbsOrigin.Z + 100));
```

After:
```cs
// Doesn't leak because the Vector3 is a struct and the pointer passed directly to native and GCd/freed by C# after
pawn.Teleport((Vector3)pawn.AbsOrigin with { Z = pawn.AbsOrigin.Z + 100 });
```

Unfortunately to ultimately fix memory leaks with regards to the schema system, it will require breaking changes so this will need to be done in a breaking 2.0 release of CS# I imagine.

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Model/CBaseEntity.cs
+++ b/managed/CounterStrikeSharp.API/Core/Model/CBaseEntity.cs
@@ -16,7 +16,7 @@ public partial class CBaseEntity
         Guard.IsValidEntity(this);
 
         if (position == null && angles == null && velocity == null)
-            throw new ArgumentNullException("No valid argument");
+            throw new ArgumentException("At least one parameter must be specified");
 
         nint _position = position?.Handle ?? 0;
         nint _angles = angles?.Handle ?? 0;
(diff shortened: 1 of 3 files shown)


roflmuffin:
@KillStr3aK

KillStr3aK:
> @KillStr3aK

this change/idea may actually reduce the number of leaks, but developers would need to explicitly cast to the struct which doesn't help existing (but abandoned) plugins, also we would need a way to tell people to actually cast their params. I'm thinking about implicit casting that would be handled by the core, just people are already instantiating the vector classes when they call the `Teleport` function so not sure what would be good there, but we can discuss it pretty sure

I'm looking forward for CS# 2.0

roflmuffin:
Yeah I thought about implicit casting, but yes it wouldn't really help anyone either because people would still be instantiating CS# Vector class.

The idea behind this isn't to fix old plugins, but allow an easier path forward for current Devs to write more performant code that would be more future proof.

PR #954, opened 2025-07-29
Additional GameData changes

K4ryuu:
Don't merge it just yet, it's WIP

### DIFF
--- a/configs/addons/counterstrikesharp/gamedata/gamedata.json
+++ b/configs/addons/counterstrikesharp/gamedata/gamedata.json
@@ -23,13 +23,13 @@
   "CCSPlayerController_ChangeTeam": {
     "offsets": {
       "windows": 100,
-      "linux": 99
+      "linux": 106
     }
   },
   "CCSPlayerController_Respawn": {
     "offsets": {
       "windows": 257,
-      "linux": 259
+      "linux": 278
     }
   },
   "CBasePlayerController_SetPawn": {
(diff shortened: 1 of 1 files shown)


xLeviNx:
You can mark this as a draft till then K4ryuu

KillStr3aK:
I think we should also generate a schema dump @roflmuffin

roflmuffin:
> I think we should also generate a schema dump @roflmuffin 

Yeah that will be easier once MM is up

crlty:
@K4ryuu 
Add in Sigs:
ClientPrint (Linux)
55 48 8D 05 ? ? ? ? 48 89 E5 41 57 41 89 F7 31 F6

K4ryuu:
I’ve verified the latest Linux binary to confirm the signature, and it has now been updated as well. The PR is no longer a WIP and should be ready to merge. However, it’s important to note that one of the entries couldn’t be found, so we might want to consider removing it. I’ve left it in my file with comments for reference. Let me know if you’d like me to remove that missing entry.

KillStr3aK:
> I’ve verified the latest Linux binary to confirm the signature, and it has now been updated as well. The PR is no longer a WIP and should be ready to merge. However, it’s important to note that one of the entries couldn’t be found, so we might want to consider removing it. I’ve left it in my file with comments for reference. Let me know if you’d like me to remove that missing entry.

"removing the missing entry" will cause loads of errors during [runtime](https://github.com/roflmuffin/CounterStrikeSharp/blob/main/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs#L33-L60), I'd assume this pr as a draft as long as we still miss the vfunc offset

PR #1007, opened 2025-08-21
Use shared libgcc and libc++

markus-wa:
This makes sure we use the same as Source2, which uses shared versions

Also resolves mismatch with custom `LD_PRELOAD` libs that use shared libcc/libc++


markus-wa:
ping @roflmuffin - is this something we can get merged? Any objections?

roflmuffin:
This change would mean you can only ever run in steamrt environment right (though you probably should be anyway)

markus-wa:
Hmm, I don't think so - my understanding is that CS2 itself is compiled like this so CS2 already uses these shared libs
This just makes it so that CSS uses the same shared libs as CS2 (whatever the ones installed on the system are) rather than statically compiled ones which may mismatch and cause conflicts with the system-installed ones

markus-wa:
Any news :eyes: - would prefer using the official releases rather than keeping a fork in sync on every update :smile: :crossed_fingers:

PR #1009, opened 2025-08-22
Small fix for new FindPickerEntity implementation

STL1te:
As I can see there is no support for nullable args in VirtualFunction.Create
So it throws "Invalid argument type(s) supplied to Virtual Function" for nullable args.

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs
+++ b/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs
@@ -39,13 +39,13 @@ public partial class CCSGameRules
 
         if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
         {
-            VirtualFunctionWithReturn<CCSGameRules, CBasePlayerController, IntPtr?, CBaseEntity?> CCSGameRules_FindPickerEntity = new(Handle, GameData.GetOffset("CCSGameRules_FindPickerEntity"));
-            return CCSGameRules_FindPickerEntity.Invoke(this, player, null);
+            VirtualFunctionWithReturn<CCSGameRules, CBasePlayerController, IntPtr, CBaseEntity> CCSGameRules_FindPickerEntity = new(Handle, GameData.GetOffset("CCSGameRules_FindPickerEntity"));
+            return CCSGameRules_FindPickerEntity.Invoke(this, player, IntPtr.Zero);
         }
         else
         {
-            VirtualFunctionWithReturn<CCSGameRules, CBasePlayerController, IntPtr?, double, double, CBaseEntity?> CCSGameRules_FindPickerEntity = new(Handle, GameData.GetOffset("CCSGameRules_FindPickerEntity"));
-            return CCSGameRules_FindPickerEntity.Invoke(this, player, null, 0, 0); // on linux we have a fourth and fifth parameter aswell, but they are unused because the condition for them is always unmet.
+            VirtualFunctionWithReturn<CCSGameRules, CBasePlayerController, IntPtr, double, double, CBaseEntity> CCSGameRules_FindPickerEntity = new(Handle, GameData.GetOffset("CCSGameRules_FindPickerEntity"));
+            return CCSGameRules_FindPickerEntity.Invoke(this, player, IntPtr.Zero, 0, 0); // on linux we have a fourth and fifth parameter aswell, but they are unused because the condition for them is always unmet.
         }
     }
 


KillStr3aK:
> As I can see there is no support for nullable args in VirtualFunction.Create So it throws "Invalid argument type(s) supplied to Virtual Function" for nullable args.

Do you have any stack where this happened?
that exception is thrown for invalid data types, not for null args and the type checking should work as intended

even though, I would agree to the changes but we should also consider my comment here: https://github.com/roflmuffin/CounterStrikeSharp/commit/245f55daf3a474bf95d0e871aa8c4117394593a3#r164365459

I'm not sure if we have to do that runtime platform check before each call as based on people's response its enough to pass 3 params to the function and it works as intended, so we might be able to just remove the 2 doubles from the linux version and use the same prototype for both

STL1te:
> Do you have any stack where this happened? that exception is thrown for invalid data types, not for null args and the type checking should work as intended

```
13:37:52 [EROR] (cssharp:Core) Error invoking callback
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Exception: Invalid argument type(s) supplied to Virtual Function
   at CounterStrikeSharp.API.Modules.Memory.VirtualFunction.Create[TArg1,TArg2,TArg3,TResult](IntPtr objectPtr, Int32 offset) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctionOffset.cs:line 391
   at CounterStrikeSharp.API.Modules.Memory.VirtualFunctionWithReturn`4..ctor(IntPtr objectPtr, Int32 offset) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctionWithReturn.cs:line 110
   at CounterStrikeSharp.API.Core.CCSGameRules.FindPickerEntityInternal(CBasePlayerController player) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs:line 42
   at CounterStrikeSharp.API.Core.CCSGameRules.FindPickerEntity[T](CBasePlayerController player) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs:line 55
   at CounterStrikeSharp.API.Core.CCSGameRules.GetClientAimTarget(CCSPlayerController player) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs:line 67
   at AllHeroes.Hero.Rebel.Hero.AhHeroRebel.GetAimTarget()
```

As I understand we have a map between csharp types and DataType in DataTypeExtensions so it cannot find an appropriate DataType for nullable args because typeof(IntPtr) != typeof(IntPtr?)

KillStr3aK:
> > Do you have any stack where this happened? that exception is thrown for invalid data types, not for null args and the type checking should work as intended
> 
> ```
> 13:37:52 [EROR] (cssharp:Core) Error invoking callback
> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
>  ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
>  ---> System.Exception: Invalid argument type(s) supplied to Virtual Function
>    at CounterStrikeSharp.API.Modules.Memory.VirtualFunction.Create[TArg1,TArg2,TArg3,TResult](IntPtr objectPtr, Int32 offset) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctionOffset.cs:line 391
>    at CounterStrikeSharp.API.Modules.Memory.VirtualFunctionWithReturn`4..ctor(IntPtr objectPtr, Int32 offset) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Modules/Memory/VirtualFunctionWithReturn.cs:line 110
>    at CounterStrikeSharp.API.Core.CCSGameRules.FindPickerEntityInternal(CBasePlayerController player) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs:line 42
>    at CounterStrikeSharp.API.Core.CCSGameRules.FindPickerEntity[T](CBasePlayerController player) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs:line 55
>    at CounterStrikeSharp.API.Core.CCSGameRules.GetClientAimTarget(CCSPlayerController player) in /home/runner/work/CounterStrikeSharp/CounterStrikeSharp/managed/CounterStrikeSharp.API/Core/Model/CCSGameRules.cs:line 67
>    at AllHeroes.Hero.Rebel.Hero.AhHeroRebel.GetAimTarget()
> ```
> 
> As I understand we have a map between csharp types and DataType in DataTypeExtensions so it cannot find an appropriate DataType for nullable args because typeof(IntPtr) != typeof(IntPtr?)

ah yeah thats right because `IntPtr` is a value type, it only works with references
thanks for the heads up!

KillStr3aK:
we could get the underlying type for nullable types before doing the data type checks to support nullable value types which might be a better idea in the long term run

STL1te:
> we could get the underlying type for nullable types before doing the data type checks to support nullable value types which might be a better idea in the long term run

I don't have sufficient knowledge about the C++ side, so I'm not sure that it can accept such changes from CSharp side without modifications. It seems that nullable types from csharp needs to be converted to some kind of default values at C++ side but this conversion can lead to undefined behavior

KillStr3aK:
> > we could get the underlying type for nullable types before doing the data type checks to support nullable value types which might be a better idea in the long term run
> 
> I don't have sufficient knowledge about the C++ side, so I'm not sure that it can accept such changes from CSharp side without modifications. It seems that nullable types from csharp needs to be converted to some kind of default values at C++ side but this conversion can lead to undefined behavior

nullable types works as intended if the underlying type is a reference type, but value types are not the same CLR types in this case, and thats why you got the exception. to make nullable value types work we could just get the underlying type and assign a data type to it

PR #1012, opened 2025-08-22
[Experimental] NuGet Dependency Resolver for Plugins

dxqwww:
Hello, let me introduce an **expiremental plugin dependency resolver** to improve how external dependencies are resolved when loading plugins.

## Problem

By default, `McMaster.NETCore.Plugins` does not handle scenarios where NuGet packages need to be loaded from a shared/global location. To address this, I implemented a custom resolver that can locate dependencies in the NuGet packages root folder (either from the `NUGET_PACKAGES` environment variable or the default `.nuget/packages` location).

## Motivation

Previously, for every plugin that depended on a NuGet package, you had to manually place the assemblies into the `shared` folder. These changes remove that friction by enabling automatic resolution.

## Configuration

Since this feature is experimental, it is disabled by default.
To enable it, set the following in `core.json`:

```json
"UseExperimentalPluginDependencyResolver": true
```

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/CoreConfig.cs
+++ b/managed/CounterStrikeSharp.API/Core/CoreConfig.cs
@@ -53,6 +53,9 @@ namespace CounterStrikeSharp.API.Core
         [JsonPropertyName("PluginAutoLoadEnabled")]
         public bool PluginAutoLoadEnabled { get; set; } = true;
 
+        [JsonPropertyName("UseExperimentalPluginDependencyResolver")]
+        public bool UseExperimentalPluginDependencyResolver { get; set; }
+
         [JsonPropertyName("ServerLanguage")]
         public string ServerLanguage { get; set; } = "en";
 
@@ -114,6 +117,8 @@ namespace CounterStrikeSharp.API.Core
         /// When enabled, plugins are automatically loaded from the plugins directory on server start.
         /// </summary>
         public static bool PluginAutoLoadEnabled => _coreConfig.PluginAutoLoadEnabled;
+
+        public static bool UseExperimentalPluginDependencyResolver => _coreConfig.UseExperimentalPluginDependencyResolver;
 
         public static string ServerLanguage => _coreConfig.ServerLanguage;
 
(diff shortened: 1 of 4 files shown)


KillStr3aK:
mind elaborate? there is already an option to put dependencies in the `shared` folder (and it is intended)

dxqshka:
> mind elaborate? there is already an option to put dependencies in the `shared` folder (and it is intended)

Yeah sure, by design plugin dependencies (like NuGet packages) can already be resolved by copying them into the `shared` folder. This works, but it requires every NuGet dependency of every plugin to be manually mirrored there, which quickly becomes tedious.

This NuGet resolver doesn't replace the `shared` folder mechanism but complements it. Instead of forcing all dependencies into `shared` manually, the resolver could automatically locate them in the system NuGet packages root. This reduces manual setup while keeping the original `shared` approach available.

dxqshka:
Here is real use case issue:

When a plugin references a NuGet package that isn't copied yet into the `shared` loading fails. The assembly is present in the local NuGet cache and is the same way we resolve `CounterStrikeSharp.API`, but any attempt to use it either from plugins or other shared projects throws at load time.

This problem is fully reproducible with `McMaster.NETCore.Plugins` library.

Similar issues:

https://github.com/natemcmaster/DotNetCorePlugins/issues/294
https://github.com/natemcmaster/DotNetCorePlugins/issues/300
https://github.com/natemcmaster/DotNetCorePlugins/issues/302

img

roflmuffin (dismissed):

roflmuffin:
Sorry for the delay in reviewing this one. Would you be able to add the new config section to the example.json and also the documentation for core config?

https://github.com/roflmuffin/CounterStrikeSharp/blob/main/configs/addons/counterstrikesharp/configs/core.example.json

dxqshka:
> Sorry for the delay in reviewing this one. Would you be able to add the new config section to the example.json and also the documentation for core config?
> 
> https://github.com/roflmuffin/CounterStrikeSharp/blob/main/configs/addons/counterstrikesharp/configs/core.example.json

Yes, sure

dxqshka:
@roflmuffin I've added some documentation for this feature. Let me know if it meets your needs

roflmuffin (approved):

roflmuffin:
@dxqwww Can you either allow contributors to edit this branch or pull latest main into your branch to allow automerge to go through?

dxqshka:
🎉

PR #1020, opened 2025-08-28
Update gamedata.json

xLeviNx:
update CCSPlayerPawnBase_PostThink linux sig

Added updates to commitsuicide, and respawn

### DIFF
--- a/configs/addons/counterstrikesharp/gamedata/gamedata.json
+++ b/configs/addons/counterstrikesharp/gamedata/gamedata.json
@@ -278,7 +278,7 @@
       "linux": 26
     }
   },
-  "ISource2GameEntities_CheckTransmit": {
+  "ISource2GameEntities::CheckTransmit": {
     "offsets": {
       "windows": 12,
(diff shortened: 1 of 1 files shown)


configs/addons/counterstrikesharp/gamedata/gamedata.json
```diff
@@ -278,7 +278,7 @@
       "linux": 26
     }
   },
-  "ISource2GameEntities::CheckTransmit": {
+  "ISource2GameEntities_CheckTransmit": {
```
ianlucas, line 281: Was this name change intended? Need to update the code if so. https://github.com/roflmuffin/CounterStrikeSharp/blob/adccc4b9e9fd5965388a8596bfe61f93c3802a4d/src/core/managers/entity_manager.cpp#L55

xLeviNx: Good catch, for some reason this was renamed from the gamedata checker

PR #1021, opened 2025-08-29
Update Utilities.cs

devayee:
Refactored and cleaned up the Utilities class to improve readability and maintainability without removing any existing functionality.


KillStr3aK (changes requested):
I don't think any of these should be changed, let me know if you have something to say about your change in `SetStateChanged`

PR #1044, opened 2025-09-19
feat: improve getplayers & RemoveItemByDesignerName

schwarper:
- feat: add CBasePlayerWeaponExtensions with GetWeaponName and GetOwner
- feat: optimize MaxPlayers by replacing Lazy<T> with static readonly int
- feat: improve GetPlayers performance and RemoveItemByDesignerName reliability
- feat: add RemoveItemBySlot

Not tested

### DIFF
--- a/managed/CounterStrikeSharp.API/Modules/Extensions/CBasePlayerWeaponExtensions.cs
+++ b/managed/CounterStrikeSharp.API/Modules/Extensions/CBasePlayerWeaponExtensions.cs
@@ -1,0 +1,48 @@
+﻿using CounterStrikeSharp.API.Modules.Entities;
+
+namespace CounterStrikeSharp.API.Modules.Extensions;
+
+public static class CBasePlayerWeaponExtensions
+{
+    /// <summary>
+    /// Gets the weapon's designer name (e.g., "weapon_ak47").
+    /// </summary>
+    /// <param name="weapon">The <see cref="CBasePlayerWeapon"/> instance.</param>
+    /// <returns>The designer name of the weapon as a string, or <c>null</c> if it cannot be retrieved.</returns>
+    public static string? GetWeaponName(this CBasePlayerWeapon weapon)
+    {
+        return weapon.GetVData<CCSWeaponBaseVData>()?.Name;
+    }
+
+    /// <summary>
+    /// Gets the owner of a weapon entity
+    /// </summary>
+    /// <param name="weapon">The weapon entity.</param>
+    /// <returns>The <see cref="CCSPlayerController"/> instance for the player, or <c>null</c> if it doesn't exist.</returns>
+    public static CCSPlayerController? GetOwner(this CBasePlayerWeapon weapon)
+    {
+        SteamID? steamId = null;
+
+        if (weapon.OriginalOwnerXuidLow > 0)
+            steamId = new(weapon.OriginalOwnerXuidLow);
+
+        CCSPlayerController? player;
+
+        if (steamId != null && steamId.IsValid())
+        {
+            player = Utilities.GetPlayerFromSteamId(steamId.SteamId64);
+
+            if (player == null)
+                player = Utilities.GetPlayerFromSteamId(weapon.OriginalOwnerXuidLow);
+        }
+        else
+        {
+            player = Utilities.GetPlayerFromIndex((int)weapon.OwnerEntity.Index);
+
+            if (player == null)
+                player = Utilities.GetPlayerFromIndex((int)weapon.As<CCSWeaponBaseGun>().OwnerEntity.Value!.Index);
+        }
+
+        return !string.IsNullOrEmpty(player?.PlayerName) ? player : null;
+    }
+}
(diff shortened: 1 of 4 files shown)


managed/CounterStrikeSharp.API/Utilities.cs
```diff
@@ -60,13 +57,19 @@ public static IEnumerable<T> FlagsToList<T>(this T flags) where T : Enum
...
-            var player = GetEntityFromIndex<CCSPlayerController>(index);
-            if (player == null || player.DesignerName != "cs_player_controller")
+            var entityPtr = EntitySystem.GetEntityByIndex((uint)index);
+            if (entityPtr is null || entityPtr == IntPtr.Zero)
+            {
+                return null;
+            }
+
+            var controller = new CCSPlayerController(entityPtr.Value);
+            if (!controller.IsValid || controller.Connected != PlayerConnectedState.PlayerConnected)
```
roflmuffin, line 67: I don't think we should move the connected state check into this method, we want this to be able to return non connected players, and only the GetPlayers util should check this.

managed/CounterStrikeSharp.API/Server.cs
```diff
@@ -142,7 +142,8 @@ public static void PrintToChatAll(string message)
 
         public static string GameDirectory => NativeAPI.GetGameDirectory();
 
-        public static int MaxPlayers => NativeAPI.GetMaxClients();
+        private static readonly int _maxPlayers = NativeAPI.GetMaxClients();
```
roflmuffin, line 145: I think this is dangerous because globals may not be ready when it runs the static initializer.  May be better to have `private static readonly int? _maxPlayers = null` and then in the accessor for the old MaxPlayers return _maxPlayers if not null, otherwise read maxplayers, set the variable and return it

managed/CounterStrikeSharp.API/Utilities.cs
```diff
@@ -81,47 +84,50 @@ public static IEnumerable<T> FlagsToList<T>(this T flags) where T : Enum
...
+                .Where(w => w?.As<CCSWeaponBase>().VData?.GearSlot == slot)
+                .FirstOrDefault();
 
-                return true;
-            }
+            if (weapon == null)
+                return false;
 
-            return false;
+            weapon.AddEntityIOEvent("Kill", weapon, delay: 0.1f);
```
roflmuffin, line 129: Why 0.1f, can this be `0.0f`?

schwarper: 0.1f is the safest for me. No crash. I never tested 0.0f

managed/CounterStrikeSharp.API/Server.cs
schwarper: Done

managed/CounterStrikeSharp.API/Utilities.cs
schwarper: Done

managed/CounterStrikeSharp.API/Modules/Extensions/CBasePlayerWeaponExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return weapon.GetVData<CCSWeaponBaseVData>()?.Name;
+    }
+
+    /// <summary>
+    /// Gets the owner of a weapon entity
+    /// </summary>
+    /// <param name="weapon">The weapon entity.</param>
+    /// <returns>The <see cref="CCSPlayerController"/> instance for the player, or <c>null</c> if it doesn't exist.</returns>
+    public static CCSPlayerController? GetOwner(this CBasePlayerWeapon weapon)
```
roflmuffin, line 22: I think this method needs to be named more clearly that it will return the econ owner not just the world owner.

managed/CounterStrikeSharp.API/Modules/Extensions/CBasePlayerWeaponExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        if (steamId != null && steamId.IsValid())
+        {
+            player = Utilities.GetPlayerFromSteamId(steamId.SteamId64);
+
+            if (player == null)
+                player = Utilities.GetPlayerFromSteamId(weapon.OriginalOwnerXuidLow);
+        }
+        else
+        {
+            player = Utilities.GetPlayerFromIndex((int)weapon.OwnerEntity.Index);
```
roflmuffin, line 40: I think this whole block could be replaced with `player = weapon.OwnerEntity.Get().As<CCSPlayerController>();`. 

It's better to use the handles `.Get()` since it will ensure that the entity is the actual same entity (and not a different player controller for instance). I don't think the `CCSWeaponBaseGun` cast below adds any value.

managed/CounterStrikeSharp.API/Modules/Extensions/CBasePlayerWeaponExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            if (player == null)
+                player = Utilities.GetPlayerFromIndex((int)weapon.As<CCSWeaponBaseGun>().OwnerEntity.Value!.Index);
+        }
+
+        return !string.IsNullOrEmpty(player?.PlayerName) ? player : null;
```
roflmuffin, line 46: Is there a better way to achieve what you are looking for here? Connection status perhaps rather than name?

roflmuffin (approved):

PR #1047, opened 2025-09-22
feat: Implement `TerminateSelf(string reason)` to allow plugins to safely terminate themselves

ELDment:
Prior to implementing `TerminateSelf`, plugins could only prevent initialization by throwing exceptions within their Load() override method. **Once fully loaded, however, plugins lacked any elegant mechanism for self-termination**

So, we implemented a state-aware TerminateSelf method:
- PluginState.Loading: Abort initialization by throwing an exception (handled by outer `try-catch`)
- PluginState.Loaded: Trigger the standard unload sequence for safe termination while preserving reload capability through css_plugins commands

image
image

**btw.** Rather than injecting the complete `PluginContext` into `BasePlugin` (which would enable **dangerous control inversion** allowing plugins to manipulate framework internals or other plugin instances), we employed interface segregation by introducing `ISelfPluginControl`. This restricted interface exposes only self-termination capabilities to plugins, **100% adhering to the principle of least privilege** and preventing unauthorized access to framework components or cross-plugin manipulation 🤗

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Application.cs
+++ b/managed/CounterStrikeSharp.API/Core/Application.cs
@@ -122,122 +122,128 @@ namespace CounterStrikeSharp.API.Core
             switch (info.GetArg(1))
             {
                 case "list":
-                {
-                    info.ReplyToCommand(
-                        $"  List of all plugins currently loaded by CounterStrikeSharp: {_pluginManager.GetLoadedPlugins().Count()} plugins loaded.");
-
-                    foreach (var plugin in _pluginManager.GetLoadedPlugins())
-                    {
-                        var sb = new StringBuilder();
-                        sb.AppendFormat("  [#{0}:{1}]: \"{2}\" ({3})", plugin.PluginId,
-                            plugin.State.ToString().ToUpper(), plugin.Plugin?.ModuleName ?? "Unknown",
-                            plugin.Plugin?.ModuleVersion ?? "Unknown");
-                        if (!string.IsNullOrEmpty(plugin.Plugin?.ModuleAuthor))
-                            sb.AppendFormat(" by {0}", plugin.Plugin.ModuleAuthor);
-                        if (!string.IsNullOrEmpty(plugin.Plugin?.ModuleDescription))
-                        {
-                            sb.Append("\n");
-                            sb.Append("    ");
-                            sb.Append(plugin.Plugin.ModuleDescription);
-                        }
-
-                        info.ReplyToCommand(sb.ToString());
-                    }
-
-                    break;
-                }
-                case "start":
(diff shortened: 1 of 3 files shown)


ELDment:
Nothing 😪

ELDment:
image

To be honest, I originally tried to recreate *SourceMod's* `SetFailState` functionality. However, since **CSS** has better practices for plugin management, I switched focus to implementing safe unloading instead (rather than putting plugins into a failed state)

Meh, I'm not entirely sure this follows best practices though, so I'd appreciate any thoughts or feedback you might have

ELDment:
### ~~[Need Help] Implementing Silent Execution Termination~~

~~I want `TerminateSelf` **to behave** like throwing an exception (clears call stack and terminates execution immediately) but **without**:~~
- ~~Printing stack traces to console~~
- ~~Requiring try-catch blocks everywhere~~

**~~Current Situation~~**:

~~We're using TerminateWithReason() in our plugins, but it doesn't provide the *"immediate termination"* behavior we need. **The call stack unwinds back to OnTestTerminateCommand, and execution continues normally until the method exits** 😅~~
image

**~~Desired Behavior~~**:

```csharp
[ConsoleCommand("Terminate")]
public void OnTestTerminateCommand(CCSPlayerController? player, CommandInfo command)
{
    Console.WriteLine("BEFORE");
    TerminateSelf("TEST");      // Should terminate silently here
    Console.WriteLine("AFTER"); // This should never execute
}
```

---

~~To prevent control flow from unwinding back to the original caller, we experimented with throwing an exception after `Unload(false);`~~
image
~~While this successfully terminates execution due to the absence of exception handlers in the call stack, it results in unhandled exception stack traces being logged to the console, **making this approach inelegant and unsuitable for production use**~~

ELDment:
Nothing 😪

ELDment:
Well, we successfully solved the ***Silent Execution Termination*** problem in commit **[ba6a8b7](https://github.com/roflmuffin/CounterStrikeSharp/pull/1047/commits/ba6a8b7c2716b58ca5784a12c98f095b357292e9)**

image

ELDment:
When `TerminateSelf` is called **within the Load override**, we only log the termination reason without stack traces

image
image


---

However, when plugins throw unexpected errors during loading (**Since this is not intentional on our part**), we provide full stack trace details since these are unintentional failures
image
image

ELDment:
Now `TerminateSelf` still works flawlessly under concurrent programming conditions 🥳
image
image

ELDment:
@roflmuffin 
**I think we're good to go now, let's do a code review 🤠**

roflmuffin (approved):

PR #1052, opened 2025-09-27
feat: add symbols support in `FindSignature`

qstage:


src/core/memory.cpp
```diff
@@ -30,5 +30,5 @@ void* FindSignature(const char* moduleName, const char* bytesStr)
         return nullptr;
     }
 
-    return module->FindSignature(bytesStr);
+    return bytesStr[0] == '@' ? module->FindSymbol(bytesStr + 1) : module->FindSignature(bytesStr);
```
roflmuffin: Won't this cause issues if the first byte of a signature is 0x40, like it currently is for `CGameEventManager_Init` on windows?

PR #1078, opened 2025-10-20
NativeBridge integration, improved vtable resolution and automatic capability reloads

SlynxCZ:
This update replaces the legacy Helpers static P/Invoke mechanism with a unified NativeBridge that dynamically loads the native CounterStrikeSharp library at runtime.
The internal FindVirtualTable resolver was upgraded to return the real class base pointer instead of RTTI, improving accuracy for low-level integrations.
Additionally, the Capabilities system now automatically unregisters and reloads all capabilities when a registrator plugin is hot-reloaded, ensuring consistent state and seamless plugin updates.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -51,8 +51,7 @@ void* GetVirtualTable(CModule* module, const std::string& name)
     // Windows RTTI format: .?AVClassName@@
     std::string decoratedTableName = ".?AV" + name + "@@";
 
-    SignatureIterator sigIt(runTimeData->m_pBase, runTimeData->m_iSize,
-                            (const byte*)decoratedTableName.c_str(),
+    SignatureIterator sigIt(runTimeData->m_pBase, runTimeData->m_iSize, (const byte*)decoratedTableName.c_str(),
                             decoratedTableName.size() + 1);
 
     void* typeDescriptor = sigIt.FindNext(false);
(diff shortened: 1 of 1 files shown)


managed/CounterStrikeSharp.API/Core/Plugin/PluginContext.cs
```diff
@@ -279,6 +279,19 @@ public void Unload(bool hotReload = false)
             }
             finally
             {
+                if (!hotReload)
+                {
+                    try
+                    {
+                        Capabilities.Capabilities.UnregisterAll(Plugin.GetType().Assembly);
```
roflmuffin: I think ideally we should know whether or not UnregisterAll unloaded anything and only display the message for "Removed all capability" if we actually removed anything, as not to clutter up the log/console if we didn't do anything (as most plugins probably don't have capabilities)

SlynxCZ:
Sure,when I will be home, I will change requested.

managed/CounterStrikeSharp.API/Core/NativeBridge.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+        catch (Exception ex)
+        {
+            _staticLogger?.LogError(ex, "Missing export {ExportName}", exportName);
+            return null;
+        }
+    }
+
+    [SecurityCritical]
+    public static void InvokeNative(IntPtr ptr)
```
roflmuffin: Can I ask the main reason for this change? Is there a performance improvement or something. While I admit the existing "Helper" class is not that well structured, this change just looks like code churn for no real reason to be honest.

SlynxCZ: Using nativelibrary class to open native library is by my sources "faster" to use as with [DllImport] every invoke gets new handle. It is better to get handle once and use it later.

SlynxCZ:
Still as I said on discord, I will close this PR and create separate ones.

SlynxCZ:
Closing this PR as separate ones will be made.

PR #1080, opened 2025-10-21
Fix: FindVirtualTable gives RTTI instead of entrypoint

SlynxCZ:
Hi, this is fix of **FindVirtualTable** method exposed to **NativeAPI**, Issue was that I accidentally used method from **CModule** to fetch vtable. I trought this method works same as DynLibUtils **GetVirtualTableByName** (nope it didnt :D). Method from CModule did gave pointer to vtables **RTTI**, this is wrong, it should give vtable **entrypoint**, as getting entrypoint from RTTI in **C#** is painful and sometimes caused me crashes when debugging with **AcceleratorCSS**. This is now fixed thanks to 2 new helper methods in **memory.cpp**👍

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -36,126 +36,13 @@ void* FindSignature(const char* moduleName, const char* bytesStr)
     return module->FindSignature(bytesStr);
 }
 
-#ifdef _WIN32
-void* GetVirtualTable(CModule* module, const std::string& name)
-{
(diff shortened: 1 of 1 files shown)


SlynxCZ:
Yeah and double checked, before logic was correct. Thank you for correcting me @Poggicek !

SlynxCZ:
Then I will create one more commit that fixes hash naming for API. That is all.

PR #1121, opened 2025-11-16
fix: Prevent VirtualFunction memory leak in native layer

ELDment:
# Prevent VirtualFunction memory leak in native layer

image

```
===== Dumping leaks =====
        Vector: 0 (0 B)
        Angles: 0 (0 B)
        GameEvents: 0 (~B)
        Virtual Functions: 2 (160 B)
        Total size: 160 B
===== Dumping leaks =====
```

**Based on** the idea from #1090 and suggestions from https://github.com/roflmuffin/CounterStrikeSharp/pull/1090#issuecomment-3449661045 and https://github.com/roflmuffin/CounterStrikeSharp/pull/1090#issuecomment-3466167683

### DIFF
--- a/configs/addons/counterstrikesharp/lang/pt-BR.json
+++ b/configs/addons/counterstrikesharp/lang/pt-BR.json
@@ -1,0 +1,27 @@
+{
+  "menu.button.previous": "Voltar",
+  "menu.button.next": "Próximo",
+  "menu.button.close": "Sair",
+
+  "all": "todos os jogadores",
+  "bots": "bots",
+  "humans": "humanos",
+  "alive": "jogadores vivos",
+  "dead": "jogadores mortos",
+  "notme": "todos os jogadores, exceto você",
+  "ct": "jogadores CT",
+  "t": "jogadores T",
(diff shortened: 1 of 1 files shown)


src/scripting/natives/natives_memory.cpp
```diff
@@ -64,12 +100,29 @@ ValveFunction* CreateVirtualFunctionBySignature(ScriptContext& script_context)
...
+        return it->second;
+    }
+
     auto function = new ValveFunction(function_addr, CONV_CDECL, args, return_type);
     function->SetSignature(signature_hex_string);
 
-    CSSHARP_CORE_TRACE("Created virtual function, pointer found at {}, signature {}", function_addr, signature_hex_string);
-
     m_managed_ptrs.push_back(function);
+    m_virtualFunctionCache[cacheKey] = function;
```
roflmuffin: If we are now storing everything in the lookup, we can probably remove `m_managed_ptrs` now

roflmuffin (approved):

roflmuffin:
Thank you!

PR #1143, opened 2025-12-11
feat(memory): allow hooking of offset based virtual functions

roflmuffin:
Thanks to @KillStr3aK for doing most of the work on this in #617


roflmuffin:
I think we can probably ship this and ignore the concerns about the duplicate name/offset pair @KillStr3aK ?

Issue #1151, opened 2025-12-16
The server is crashing


Tajemniq:
I just join the server, try to change the mode with the !modes command, and suddenly after a moment the server shuts down.

Server logs: https://pastebin.com/qfTichB4

geqo:
Same for me, after any shot from weapon I got crash with this:
ResetBreakpadAppId: Universe is 1 (k_EUniversePublic)
ResetBreakpadAppId: Setting dedicated server app id: 2347773

roflmuffin:
The change in 1.0.352 does not seem to be compatible with CS2Fixes. I'm investigating if there is a way forward to make it still compatible, but I'd advise reverting until then.

qsxz25:
> Same for me, after any shot from weapon I got crash with this: ResetBreakpadAppId: Universe is 1 (k_EUniversePublic) ResetBreakpadAppId: Setting dedicated server app id: 2347773

+1, same with WeaponPaints plugin

Tajemniq:
I rolled back the version and I have another problem with the menu. I have commands such as !modes, !maps, and !settings on my server, and the menu won't appear in the middle of the screen. I don't know how to fix it.

roflmuffin:
Please try on latest v353

sunrisecz:
> Please try on latest v353

v353 is incompatible with CS2Fixes. Server crashes.

Tajemniq:
> Please try on latest v353

After updating to version v353, the menus are still not displayed

sunrisecz:
> > Please try on latest v353
> 
> v353 is incompatible with CS2Fixes. Server crashes.

v354 is incompatible with CS2Fixes again, and the server crashes.
OS:windows

sunrisecz:
> > > Please try on latest v353
> > 
> > 
> > v353 is incompatible with CS2Fixes. Server crashes.
> 
> v354 is incompatible with CS2Fixes again, and the server crashes. OS:windows

v1.0.355 is now works!

Oz-Lin:
v355 crashes on Linux Ubuntu 24.04, CS2Fixes + CS# (or CS# + SwiftlyS2 for those who are transitioning plugin framework ecosystems)

strenx22:
> v355 crashes on Linux Ubuntu 24.04, CS2Fixes + CS# (or CS# + SwiftlyS2 for those who are transitioning plugin framework ecosystems)

did you solve the problem somehow?

in particular, Sharp + Swift2

Issue #1148, opened 2025-12-15
v1.0.352 update  windows server,  crashes on startup


sunrisecz:

FPTyel:
Debian 13 stable release with latest updates, also crashes on startup

sunrisecz:
v1.0.355 is now works!

PR #1163, opened 2025-12-27
fix: Ensure compatibility with `CS2Fixes`

ELDment:
#1151 
https://github.com/roflmuffin/CounterStrikeSharp/issues/1151#issuecomment-3688293977

**this pr fixes the crash on windows at least,** linux is untested

00145a87-3fdc-48c4-b2ed-28aa66fdaddd

### Root cause:

CS# hooks `CBaseEntity_TakeDamageOld` at startup, but CS2Fixes needs to hook it too, since CS# loads before CS2Fixes, once CS# hooks the function, CS2Fixes can no longer find the original signature in memory (leading to errors)

### Fix:

we now only cache the memory address of `CBaseEntity_TakeDamageOld` during CS# startup, then **defer the actual hook until all Metamod plugins have finished loading**

### DIFF
--- a/src/core/managers/entity_manager.cpp
+++ b/src/core/managers/entity_manager.cpp
@@ -15,7 +15,6 @@
  */
 
 #include "core/managers/entity_manager.h"
-// #include "core/detours.h"
 #include "core/function.h"
 #include "core/gameconfig.h"
 #include "core/globals.h"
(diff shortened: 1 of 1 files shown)


roflmuffin:
This looks good to me. Could we remove the commented out code from the old file and then I'll merge?

ELDment:
> This looks good to me. Could we remove the commented out code from the old file and then I'll merge?

updated

ELDment:
### Side note:

possibly related to #1148 as well, not sure

roflmuffin (approved):

PR #1182, opened 2026-01-20
refactor: Fix memory leak caused by allocating `Vector`, `QAngle`, etc. class objects

ELDment:
```csharp
        {
            RegisterListener<Listeners.OnTick>(() =>
            {
                for (int j = 0; j < 32; j++)
                {
                    for (int i = 0; i < 64; i++)
                    {
                        var vector = new Vector(i, i, i);
                    }
                }
            });
        }

        [ConsoleCommand("css_tp")]
        public void OnTeleportCommand(CCSPlayerController? player, CommandInfo commandInfo)
        {
            var vector = new Vector(0, 0, 500);
            commandInfo.ReplyToCommand($"{vector} -> ({NativeAPI.VectorGetX(vector.Handle)}, {NativeAPI.VectorGetY(vector.Handle)}, {NativeAPI.VectorGetZ(vector.Handle)}) | {NativeAPI.VectorIsZero(vector.Handle)}");
            NativeAPI.VectorSetZ(vector.Handle, 1000);
            player!.PlayerPawn.Value!.Teleport(vector);
        }
```
image

Performs well on Windows with no noticeable memory leaks or CPU overhead increase
Backward compatibility tested with no issues

---

Maybe this is just a draft? i'm not sure, there might be a better way, idk
If it is good to go, please merge, lol

### DIFF
--- a/managed/CounterStrikeSharp.API/NativeHandleTracker.cs
+++ b/managed/CounterStrikeSharp.API/NativeHandleTracker.cs
@@ -5,7 +5,7 @@ namespace CounterStrikeSharp.API
 {
     internal static class NativeHandleTracker
     {
-        private sealed class Entry
+        internal sealed class Entry
         {
             public WeakReference<NativeObject> Target { get; }
             public IntPtr Handle { get; }
@@ -17,7 +17,7 @@ namespace CounterStrikeSharp.API
             }
         }
 
-        private static readonly List<Entry> _entries = new();
+        internal static readonly List<Entry> _entries = new();
         private static readonly object _lockObj = new();
         private static int _nextCleanupIndex;
         private static Timer? _timer;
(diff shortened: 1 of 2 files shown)


roflmuffin:
I think this concept seems sound to me, converting to handles only as required. I'll just run some tests on my end.

roflmuffin:
Happy to merge and release @ELDment?

ELDment:
> Happy to merge and release @ELDment?

of course 🎉

KillStr3aK:
its been a while for me, layout-wise it might be okay but wouldn't there be issues if the variable isn't allocated using memalloc?

ELDment:
> its been a while for me, layout-wise it might be okay but wouldn't there be issues if the variable isn't allocated using memalloc?

imo, we don’t need `memalloc` here because we’re not creating an engine-owned C++ object that the engine will later free via its allocator, we’re just passing a pointer to a POD layout (3 contiguous floats) that native code **reads/writes** as `Vector*` and does not **delete/free**

in that case the allocator is irrelevant, `AllocHGlobal` is fine as long as the layout matches and the memory stays alive for the duration of the call

not sure

PR #1188, opened 2026-01-23
fix: remove entity listener hack & add tests

toorisrael:


toorisrael:
sry missclick, idk how to remove it

PR #1195, opened 2026-01-23
Fix old suppliers lingering when Capability is registered on plugin reload.

killaketa:
This simple change fixes a bug in the Capabilities system. When reloaded a plugin re-registering its capability will have its old supplier provided instead of the one registered by the plugin on reload. This causes plugins using the Core plugins API to use the old supplier instead of the new one, causing a desync between the Core and the plugins trying to use it.

Only the first supplier was used when a plugin calls `Get` to retrieve the supplier so the old supplier, leaving the new supplier there but unused. The change just clears the Capability's supplier list when reassigning a new supplier.

Test latest and this PR with this plugin to see the fix tested:
https://github.com/killaketa/CS2-PluginReloadSTypesTest

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## What's Changed in v1.0.359
+* fix: update set state changed offset for linux ([3262827](https://github.com/roflmuffin/CounterStrikeSharp/commit/326282757fe094cc2eabfcb32fc8e86f6db9aabe))
+
 ## What's Changed in v1.0.358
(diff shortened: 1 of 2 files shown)


roflmuffin:
Because we only return the output of the first supplier function in player capabilities, maybe we should just replace the dictionary in those classes with a simple variable, and just set that instead of the clear and add?

killaketa:
> Because we only return the output of the first supplier function in player capabilities, maybe we should just replace the dictionary in those classes with a simple variable, and just set that instead of the clear and add?

That would honestly be the ideal solution and what I thought of first, but someone in the server told me this structure was to conform to standard dotnet practice. I'll open up a PR for this later today when I have some more time, closing this one for now.

Issue #1239, opened 2026-03-10
Subject: Server crashes on map change with full players despite tv_autorecord 0 and tv_delay 0


a4594865-crypto:
Hi,

I am experiencing a critical crashing issue with my CS2 server during map changes. Here are the details:

The Problem:
The server consistently crashes (hangs/closes) during the transition to the next map after a vote ends, but only when the server is full.

My Configurations:
I have already tried the following settings to mitigate the issue, but the crash still occurs:

tv_enable 1

tv_autorecord 0 (I trigger recording via plugins/manually)

tv_delay 0

tv_delaymapchange 1

Observation:
It seems that when the server is at max capacity, the process of disconnecting all players while simultaneously handling the CSTV shutdown/handoff causes a deadlock. The crash does not seem to happen if the server is nearly empty.

CSS-V1.0.363

Is there any known workaround for this "Full Server Map Change Crash"? Or is there a way via CSS to force a slight delay before changelevel to let the server clear the player sessions and CSTV buffer more gracefully?

Thank you for your help!

rodrigolaca:
i'm facing the same error here. @roflmuffin when you have some time, would you mind taking a look at it? seen multiple people with this same error so far. thanks!

vindict6:
I am also facing this issue. The server crashes if tv_record is set to true. The crash only happens after the vote cards, and hangs up before the loading screen appears for the next map. The end screen disappears, and it hangs on the first-person view of the current map for about ten seconds, then the server crashes.

a4594865-crypto:
Whether running CSS in vanilla mode (no plugins) or with all plugins enabled, the server randomly crashes during map transitions whenever tv_enable is set to 1. This issue is especially prominent when there are 8 or more players on the server (high occupancy). Even without active recording, the crash still occurs during the ChangeLevel process.

daffyyyy:
> Hi,
> 
> I am experiencing a critical crashing issue with my CS2 server during map transitions, and I've identified a specific trigger related to the player count.
> 
> The Problem: The server consistently crashes (hangs or closes) during the transition to the next map after a vote ends, but the behavior depends on the number of players:
> 
> 3-4 Players: The map change works fine and smoothly.
> 
> 10-11 Players (Full 5v5): The server consistently crashes during the changelevel process.
> 
> My Setup:
> 
> Framework: CounterStrikeSharp (v1.0.363)
> 
[...]

PR #1314, opened 2026-05-15
fix: guard Hook_StartupServer against spurious second call (HLTV wakeup / ss_dead)

### LINKED ISSUE #946: Server crashes with tv_broadcast 1 set and map change
If you are running a server with CounerStrikeSharp and set "tv_broadcast 1" when you then change map using the changelevel command in console (or by using Server.ExecuteCommand to run changelevel the server crashes.

I tried this with CSS off and the map changes fine, if CSS is then enabled the server will then crash when chaging map. (I removed my plugin and just had CSS run with no plugins)

Thanks

stephanebruckert:
## Problem

`Hook_StartupServer` can fire twice for the same session when CS2
internally reloads without deactivating the loop mode. The second call
fires `OnLevelEnd` → PlayerManager disconnects connected players in CSS
state → stale .NET callbacks → SEGV on the next `DispatchConCommand`.

Two known triggers:

- **Workshop map load** (`host_workshop_map`) — CS2 goes through an
  internal `ss_dead` reload to fetch the addon and fires
  `Hook_StartupServer` a second time.
- **HLTV wakeup** — when the first player connects to a hibernating
  server, HLTV starts its own session and triggers a second
  `Hook_StartupServer`.

## Fix

Track whether the call follows a genuine `OnLevelShutdown` (set in the
`OnLevelShutdown` hook). `OnStartupServer` takes a `bool levelShutdown`:

- `true` → fire `OnLevelEnd` and reset per-map tick state (genuine path)
- `false` → reset tick state only, no `OnLevelEnd` (the spurious cycle)

Tick-state reset is unconditional — `OnGameFrame`'s `universal_time`
math needs it false on the first frame of every session regardless of
which path started it.

## Testing

Verified on CS2 1.41.6.2 across 40+ map transitions: `host_workshop_map`
cycles, `changelevel` with `tv_broadcast` on, HLTV wakeup from
hibernation, MatchZy loaded throughout. No crashes.

## Related

Fixes #946 and #1239.

Also resolves the following MatchZy issues (same root cause, no MatchZy changes needed):
- https://github.com/shobhit-pathak/MatchZy/issues/380
- https://github.com/shobhit-pathak/MatchZy/issues/241
- https://github.com/shobhit-pathak/MatchZy/issues/311

### DIFF
--- a/src/mm_plugin.cpp
+++ b/src/mm_plugin.cpp
@@ -204,12 +204,28 @@ bool CounterStrikeSharpMMPlugin::Load(PluginId id, ISmmAPI* ismm, char* error, s
     return true;
 }
 
+static bool s_bLevelShutdownOccurred = false;
+
 void CounterStrikeSharpMMPlugin::Hook_StartupServer(const GameSessionConfiguration_t& config, ISource2WorldSession*, const char*)
 {
     globals::entitySystem = interfaces::pGameResourceServiceServer->GetGameEntitySystem();
+    // Remove before adding to prevent double-registration when workshop addon changes
+    // trigger a second StartupServer within the same map session (ss_dead cycle).
+    globals::entitySystem->RemoveListenerEntity(&globals::entityManager.entityListener);
     globals::entitySystem->AddListenerEntity(&globals::entityManager.entityListener);
 
-    globals::timerSystem.OnStartupServer();
+    // Only fire OnLevelEnd lifecycle after a genuine loop-mode deactivation (OnLevelShutdown).
+    // Workshop addon changes trigger a second StartupServer via an internal ss_dead cycle
+    // without deactivating the loop mode; calling OnStartupServer here fires OnLevelEnd →
+    // PlayerManager disconnects still-connected players in CSS state → stale .NET callbacks
+    // → SEGV on the next DispatchConCommand hook.
+    // OnLevelShutdown (via ILoopMode::LoopShutdown post-hook in metamod-source) fires only
+    // on genuine changelevel/shutdown, not during the ss_dead reload cycle.
+    if (s_bLevelShutdownOccurred)
+    {
+        s_bLevelShutdownOccurred = false;
+        globals::timerSystem.OnStartupServer();
+    }
 
     on_activate_callback->ScriptContext().Reset();
     on_activate_callback->ScriptContext().Push(globals::getGlobalVars()->mapname.ToCStr());
@@ -299,7 +315,7 @@ int CounterStrikeSharpMMPlugin::Hook_LoadEventsFromFile(const char* filename, bo
     RETURN_META_VALUE(MRES_IGNORED, 0);
 }
 
-void CounterStrikeSharpMMPlugin::OnLevelShutdown() {}
+void CounterStrikeSharpMMPlugin::OnLevelShutdown() { s_bLevelShutdownOccurred = true; }
 
 bool CounterStrikeSharpMMPlugin::Pause(char* error, size_t maxlen) { return true; }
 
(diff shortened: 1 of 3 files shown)


src/mm_plugin.cpp
```diff
@@ -204,12 +204,28 @@ bool CounterStrikeSharpMMPlugin::Load(PluginId id, ISmmAPI* ismm, char* error, s
...
+
 void CounterStrikeSharpMMPlugin::Hook_StartupServer(const GameSessionConfiguration_t& config, ISource2WorldSession*, const char*)
 {
     globals::entitySystem = interfaces::pGameResourceServiceServer->GetGameEntitySystem();
+    // Remove before adding to prevent double-registration when workshop addon changes
+    // trigger a second StartupServer within the same map session (ss_dead cycle).
+    globals::entitySystem->RemoveListenerEntity(&globals::entityManager.entityListener);
     globals::entitySystem->AddListenerEntity(&globals::entityManager.entityListener);
 
-    globals::timerSystem.OnStartupServer();
```
roflmuffin, line 212: Will this now no longer fire on first startup since the bool is false by default?

stephanebruckert: It still fires on first startup, but via `OnLevelInit` rather than `Hook_StartupServer`, so the flag being `false` by default is intentional. `Hook_StartupServer` only triggers after the initial load (on genuine changelevel or the spurious HLTV/ss_dead calls this fix guards against).

stephanebruckert:
Hey @roflmuffin I'm investigating a regression, please consider this PR as draft for the time being. This needs much more testing

src/mm_plugin.cpp
stephanebruckert: Updating this since the code has changed. While testing the original gated version on a live server, I hit a mid-match MatchZy failure. Tracing it, I found the `if (s_bLevelShutdownOccurred) { OnStartupServer(); }` gate was also skipping the `m_has_map_ticked` reset, which broke `universal_time` math in `OnGameFrame` and made MatchZy's AutoStart timer fire late mid-match.

So the shape is different now. `OnStartupServer(bool levelShutdown)` gets called every time, and the bool just controls whether OnLevelEnd fires. On first startup the bool is false, so OnLevelEnd doesn't fire (which was your concern). The function still runs, but the only thing it does is reset `m_has_map_ticked` / `m_has_map_simulated` to false, and they're already false at that point, so it's a no-op in practice.

stephanebruckert:
Investigation is done, this is ready for review again. Details on the code shape change are in the inline thread above, full bug and fix story in the PR description.

Validated on CS2 1.41.6.2 across 40+ map transitions: `host_workshop_map` cycles, HLTV wakeup from hibernation, `changelevel` with `tv_broadcast` on, MatchZy loaded throughout. No crashes.

If anyone wants to try the fix on their own server before merge, prebuilt binaries are at https://github.com/2hours-gg/CounterStrikeSharp/releases/tag/v1.0.368-2hours.2.

I also have unit tests for the `OnStartupServer(bool)` contract on a sibling branch ([diff](https://github.com/2hours-gg/CounterStrikeSharp/compare/fix/double-startupserver-crash...fix/double-startupserver-crash-with-tests)). Happy to either fold them into this PR or open as a follow-up, whichever you'd prefer.

roflmuffin:
Hey mate, I think there is a genuine fix in this PR but it's honestly too hard to tell what it's trying to achieve purely from the AI comments and AI generated PR description. 

I won't be continuing to review this PR, but if you want to put into a few (of your own) words the clear repro steps and the proposed fix I can look into implementing it

xiaoyueyoqwq:
Agree, this patch works for me too.

PR #1322, opened 2026-05-28
chore: upgrade to .NET 10 (LTS)

Mesharsky:
Moves CounterStrikeSharp from .NET 8 to .NET 10 with the smallest possible change surface. The public API is unchanged, the managed build has zero source breaks (ApiCompat baseline still passes), and existing .NET 8 plugins keep working without recompilation.


roflmuffin (approved):

roflmuffin:
Thank you!

PR #1331, opened 2026-06-02
update: add Ray/Hull Trace API (INavPhysicsInterface)

SlynxCZ:
Adds a complete ray and hull trace system to CounterStrikeSharp, exposing CS2's `CNavPhysicsInterface` vtable to C# plugins.

### What changed

**C++ (`src/`)**
- `src/core/cs2_sdk/interfaces/navphysicsinterface.h/.cpp` — adapts the existing `INavPhysicsInterface` static wrapper to use CSSharp's own `modules::server->FindVirtualTable()` instead of Source2Toolkit's `DynLibUtils`
- `src/scripting/natives/natives_raytrace.cpp/.yaml` — 6 new natives:
  - `TRACE_SHAPE` / `TRACE_END_SHAPE` / `TRACE_HULL_SHAPE` — write into a caller-provided `CSSTraceResult` buffer (no heap allocation on C++ side)
  - `POINT_CONTENTS` — contents bitmask at world position
  - `CHECK_AREA_OVERLAPPING_ENTITY` — nav area vs entity AABB overlap
  - `GET_ENTITY_WORLD_SPACE_AABB` — world-space AABB of an entity
- `CTraceFilter` is constructed with the entity's owner handle and hierarchy ID (read via schema with static caching), matching `CTraceFilterEx` behaviour from Source2Toolkit

**C# (`managed/`)**
- `Modules/Utils/Trace.cs`
  - `TraceOptions` — `InteractsAs`, `InteractsWith`, `InteractsExclude`
  - `TraceResult` — `[StructLayout(Sequential, Pack=1)]` unmanaged struct, stack-allocated by each `Trace.*` method and returned by value; zero heap allocation, zero native calls for field access
  - `Trace` static class — `TraceShape`, `TraceEndShape`, `TraceHullShape`, `PointContents`, `CheckAreaOverlappingEntity`, `GetEntityWorldSpaceAABB`
- `Modules/Utils/ContentsMask.cs` — `Contents` and `Mask` static classes with all `CONTENTS_*` and `MASK_*` constants from `bspflags.h` / `const.h` as `const ulong`

### Usage

```csharp
// Line trace from player eye position
var result = Trace.TraceShape(player.AbsOrigin!, player.EyeAngles, player);
if (result.DidHit())
    Server.PrintToConsole($"Hit {result.HitEntity().DesignerName} at {result.HitPoint}");

// Hull trace with custom mask
var result = Trace.TraceHullShape(start, end, mins, maxs,
    options: new TraceOptions { InteractsWith = Mask.PlayerSolid });

// AABB
Trace.GetEntityWorldSpaceAABB(entity, out var mins, out var maxs);

// Contents check
bool isSolid = (Trace.PointContents(pos) & Contents.Solid) != 0;
```

### Notes

- `gamedata.json` has no new entries required — vtable is found by RTTI name (`"CNavPhysicsInterface"`)
- `CSSTraceResult` layout is `#pragma pack(1)` and must stay in sync with `TraceResult`'s `[StructLayout(Sequential, Pack=1)]`
- `CTransform` / `RnCollisionAttr_t` are intentionally not exposed on the C# side; opaque pointers (`nint`) are available via `Body()`, `Shape()`, `Surface()`, `Hitbox()` for interop


roflmuffin:
I think this looks passable to me. It is a bit annoying with the opaque pointers to the structs we don't have definitions for, but it is better than nothing since CS#1.0 native struct support is terrible. I'll probably write some integration/game tests for this to try it out which may take me some time.

managed/CounterStrikeSharp.API/Modules/Utils/Trace.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    public Vector StartPos => new(m_vStartX, m_vStartY, m_vStartZ);
+    public Vector EndPos => new(m_vEndX, m_vEndY, m_vEndZ);
+    public Vector Normal => new(m_vHitNormalX, m_vHitNormalY, m_vHitNormalZ);
+    public Vector HitPoint => new(m_vHitPointX, m_vHitPointY, m_vHitPointZ);
+
+    public CEntityInstance HitEntity() => new(m_pEnt);
+
+    /// <summary>Native <c>CHitBox*</c> — opaque pointer.</summary>
+    public nint Hitbox() => m_pHitbox;
```
roflmuffin, line 82: It might be worth considering naming these something else, like `HitboxPtr`, just in case we do ever add the proper struct types for these later down the line and then we would have to make a breaking change to the existing property. Don't know what your thoughts are on that?

SlynxCZ: I thought about naming them the way they those members are named as we could use NativeObject and construct them in same way as CEntityInstance wrapper is. So naming would be ideal, otherwise it can be changed right away.

SlynxCZ:
> I think this looks passable to me. It is a bit annoying with the opaque pointers to the structs we don't have definitions for, but it is better than nothing since CS#1.0 native struct support is terrible. I'll probably write some integration/game tests for this to try it out which may take me some time.

Thanks, yeah I didn't imlement them as we can use NativeObject or raw unsafe / marshal passing, also I had idea of having raytrace logic in core so we can just bump SDK and its fixed right away,

managed/CounterStrikeSharp.API/Generated/Natives/API.cs
```diff
@@ -1692,26 +1799,6 @@ public static void SchemaNetworkStateChanged(IntPtr instance, uint offset, uint
...
-        public static bool IsServerPaused(){
-			lock (ScriptContext.GlobalScriptContext.Lock) {
-			ScriptContext.GlobalScriptContext.Reset();
-			ScriptContext.GlobalScriptContext.SetIdentifier(0xB216AAAC);
-			ScriptContext.GlobalScriptContext.Invoke();
-			ScriptContext.GlobalScriptContext.CheckErrors();
-			return ScriptContext.GlobalScriptContext.GetResultPrimitive<bool>();
-			}
-		}
-
```
KillStr3aK, line 1714: [these do still exist in the yaml](https://github.com/roflmuffin/CounterStrikeSharp/blob/main/src/scripting/natives/natives_server.yaml) so next code gen would re add these, what was the point of this removal?

roflmuffin: Looks like they've just been moved? I'm assuming the generator may have done this? If not they should probably be moved back

SlynxCZ: I've regenerated whole API.cs using CodeGen, basically I removed everything and ran CodeGen to have up-to-date methods

roflmuffin:
Hey @SlynxCZ have you got any examples of `Trace.PointContents(pos)` working? Unless I am misunderstood, I am checking a position where a world brush is returning on a trace but it seems to always return an empty contents mask.

managed/CounterStrikeSharp.API/Modules/Utils/Trace.cs
```diff
@@ -0,0 +1,154 @@
...
+
+    /// <summary>Returns the contents bitmask at the given world position.</summary>
+    public static Contents PointContents(Vector pos, Contents contentsMask = Contents.Empty)
+        => (Contents)NativeAPI.PointContents(pos.Handle, (ulong)contentsMask);
+
+    /// <summary>
+    /// Returns whether a nav area overlaps with an entity's bounding box.
+    /// <paramref name="area"/> is an opaque pointer to a <c>CNavArea</c> obtained from the nav system.
+    /// </summary>
+    public static bool CheckAreaOverlappingEntity(nint area, CBaseEntity entity, bool extrudeHullHeight = false)
```
roflmuffin: I would be tempted to not expose this method to the API until we have a way to get access to `CNavArea`s.

SlynxCZ:
> Hey @SlynxCZ have you got any examples of `Trace.PointContents(pos)` working? Unless I am misunderstood, I am checking a position where a world brush is returning on a trace but it seems to always return an empty contents mask.

Hi, unfortunately I don't have, I've never been in a scenario when I needed those as my plugins are based purely on TraceShape, I've just implemented them to expose full `CNavPhysicsInterface` vtable

roflmuffin (approved):

Issue #1347, opened 2026-07-09
for the dev build/pull request


zdomain23-oss:
the host signatures and other signatures are missing so plugins and commands are not fully working

Issue #1343, opened 2026-07-09
CSSHARP NOT LOADING AFTER UPDATE


MasterTishere:
METAMOD UPDATED Please update CSSHARP Too its not loading anymore after the update

veramellio:
duplicate of #1342

MasterTishere:
THANKS FOR THE UPDATE <3 lOVE YOU guyz

Issue #1342, opened 2026-07-09
Latest CS2 Update Start Error


zZaNdrei7:
Plugin won't start on the latest CS2 Update (using latest MetaMod version fixed for this update an hour ago).

counterstrikesharp.so: undefined symbol: g_bUpdateStringTokenDatabase

MasterTishere:
yes the plugin is not loading anymore .......after new update  ........they updated metamod its working but cssharp is not loading

Berusxoxo:
Metamod is loading but the cssharp is showing error after the release of Season 5 in cs2

mustafalperen:
Same, bump.

ad1k4h:
up

a4594865-crypto:
Same

Berusxoxo:
@roflmuffin sir,
Need an updated release.

wickedroid:
meta list
Listing 1 plugin:
  [01] <ERROR>
meta unload 1
Plugin 1 unloaded.
meta load addons/counterstrikesharp/bin/linuxsteamrt64/counterstrikesharp
Failed to load plugin addons/counterstrikesharp/bin/linuxsteamrt64/counterstrikesharp (/home/container/game/csgo/addons/counterstrikesharp/bin/linuxsteamrt64/counterstrikesharp.so: undefined symbol: g_bUpdateStringTokenDatabase

dRzzz-zzz:
+up

wickedroid:
don't try this build . is not working - https://github.com/roflmuffin/CounterStrikeSharp/actions/runs/29021820092

zZaNdrei7:
Did anyone manage to find perhaps a working build until the official one is done? 🙏

wickedroid:
https://github.com/roflmuffin/CounterStrikeSharp/actions/runs/29030284789/artifacts/8203602743 try this

zZaNdrei7:
> https://github.com/roflmuffin/CounterStrikeSharp/actions/runs/29030284789/artifacts/8203602743 try this 

Link not working

bonkboykz:
So, I manged to get it to run (haven't tested the games though) with this one https://github.com/bonkboykz/CounterStrikeSharp/actions/runs/29026374138/artifacts/8201873146  but there are signatures missing, so things like commands in chat (.ready etc.) will not work, will try to reverse engineer that https://github.com/ianlucas/cs2-signatures

FRHOME:
[...]

PR #1348, opened 2026-07-09
fix: Update for latest CS2 Game Update [1.41.6.9]

ELDment:
#1342 
#1343 
some changes overlap with #1344 
#1347 
perhaps related to #1351 
#1352


ELDment:
The build from https://github.com/roflmuffin/CounterStrikeSharp/actions/runs/29033402794 should resolve most known issues. I’m still checking for any remaining unknown issues

murillolosano19:
Thanks @ELDment, it worked on my server.

htbarbieri:
Thanks for this fix @ELDment — the build from the CI run resolves the `undefined symbol: g_bUpdateStringTokenDatabase` error and gets the server loading again.

After testing the `PullRequest1348.5` build, we found two remaining issues that seem to be caused by signatures **not updated in this PR**:

**1. Server crash when players connect**
The crash appears to happen at player connection time. `CBaseEntity_TakeDamageOld` was not changed in this PR — its current Linux signature (`55 48 89 E5 41 57 41 56 41 55 49 89 FD 31 FF`) may still be stale for Season 5, and since it is called frequently during gameplay it is a likely crash candidate.

**2. Knife round `.stay`/`.switch` not working + chat commands generally broken**
Chat command dispatch goes through `Host_Say`. The current Linux signature (`55 48 89 E5 41 57 49 89 F7 41 56 41 55 41 54 4D 89 C4`) was also not touched in this PR — if it is stale the whole chat command system fails silently (no error at startup, commands simply do not fire). This also breaks round backups since `.stop` relies on the same dispatch.

Both are confirmed untouched in the diff. Would be great if someone with access to the Season 5 `server.so` can run `eng/test-gamedata-signatures.py` against `Host_Say` and `CBaseEntity_TakeDamageOld` specifically.

htbarbieri:
**Follow-up after testing `PullRequest1348.5` on a live server:**

Confirmed behavior with this build:

✅ Server loads without the `undefined symbol` error  
✅ Players can connect without crash  
✅ Console commands work (`css_forcestart`, `css_fp`, `css_endmatch`, etc.)  
❌ **All chat commands are broken** (`.ready`, `.fp`, `.tech`, etc. — no response in chat)

This confirms the `Host_Say` signature is still stale. Chat commands in CSSharp plugins use `AddCommandListener("say", ...)` which hooks through `Host_Say`. Console commands (`css_*`) go through the engine ConCommand system and are unaffected.

So the two signatures from my previous comment are still the blockers:
- `Host_Say` → chat commands broken (all plugins affected)
- `CBaseEntity_TakeDamageOld` → still unverified, worth checking

The server is usable as a workaround (console RCON commands work), but any plugin relying on chat input is non-functional.

htbarbieri:
**Additional evidence — server log confirming `Host_Say` hook is not firing:**

```
[All Chat][hrK]: !ready
L 07/10/2026 - 00:08:32: "hrK<2><[U:1:...]><TERRORIST>" say "!ready"

[All Chat][hrK]: .ready  
L 07/10/2026 - 00:08:43: "hrK<2><[U:1:...]><TERRORIST>" say ".ready"
```

The game engine logs the `say` events, but CSSharp's `AddCommandListener("say", ...)` never fires — no plugin response, nothing in the CSSharp log. The hook is registered but `Host_Say` is not being found/patched, so the listener is never triggered.

This is the exact signature that needs updating to unblock all chat-based commands for every plugin running on this build.

ehwhattaugonnado:
**Six Linux vtable offsets also changed in this update — not covered by this PR**

Ran the `PullRequest1348.5`-era signatures on a live Linux match server and still got instant server crashes at match start / player connect (MatchZy loaded). Signatures all check out against the current `server.so` — the crash is from stale **offsets**, which this PR doesn't touch. Likely also the cause of the weapon allocator weirdness in #1351.

```json
"CCSPlayerController_Respawn": 272,
"CCSPlayer_ItemServices_GiveNamedItem": 24,
"CCSPlayer_ItemServices_DropActivePlayerWeapon": 27,
"CCSPlayer_ItemServices_RemoveWeapons": 28,
"CGameSceneNode_GetSkeletonInstance": 13,
"CBasePlayerPawn_CommitSuicide": 384
```

(previously 275 / 21 / 24 / 25 / 10 / 390)

Verified `GiveNamedItem` = 24 directly against `libserver.so` (depot 2347773, manifest 5502194087696430282) by walking the `CCSPlayer_ItemServices` vtable via RTTI — slot 24 tail-calls the same function this gamedata's `GiveNamedItem` signature finds; slot 21 no longer does. The rest are from the gamedata-hunting community, with reports that the respawn crash disappears once applied.

❌ Windows offsets for this update — still unverified, someone with a Windows dump would need to check.

ELDment:
@ehwhattaugonnado 
thanks for the great advice! I just installed `Matchzy` on my test server and am checking the impact now

ELDment:
feedback from the community has been great so far

we should have fixed all known and reported issues. currently, we're just digging around for any potential crashes

anik6153:
Hi I am basically new to github. So I just wanted to ask that you are going to make a updated release or do I need to download different file from different places?

roflmuffin:
Hey @ELDment have you managed to run the native game tests plugin against a new server yet? Given I think I saw some changes to it I'm imagining so. Otherwise I'll go ahead and run through it in the next couple of hours

ELDment:
> Hey @ELDment have you managed to run the native game tests plugin against a new server yet? Given I think I saw some changes to it I'm imagining so. Otherwise I'll go ahead and run through it in the next couple of hours

heya, @roflmuffin 
`TestPlugin` seems perfectly fine at runtime. But `CounterStrikeSharp.API.Tests` (NativeTestsPlugin) hit some **FAILs**. Digging into it now

roflmuffin:
It does require a specific game mode/server setting so that can also contribute to fails, but I don't have the settings on hand right now

ELDment:
> Hi I am basically new to github. So I just wanted to ask that you are going to make a updated release or do I need to download different file from different places?

heya, @anik6153 
you can refer to https://github.com/roflmuffin/CounterStrikeSharp/issues/1342#issuecomment-4932139121 and https://github.com/roflmuffin/CounterStrikeSharp/issues/1342#issuecomment-4932748065
that said, i still think there are some underlying issues hidden under the surface

ELDment:
> `TestPlugin` seems perfectly fine at runtime. But `CounterStrikeSharp.API.Tests` (NativeTestsPlugin) hit some **FAILs**. Digging into it now

@roflmuffin 
other than some minor failures within `CounterStrikeSharp.API.Tests` (NativeTestsPlugin)—particularly these types of outdated hardcoded checks like `Assert.Equal(1456, offset)`

all other tests passed. it seems there's nothing else worth worrying about

eng/test-gamedata-signatures.py
```diff
@@ -0,0 +1,306 @@
+#!/usr/bin/env python3
+import importlib
```
roflmuffin: If it's fine with you; I'd rather not include this in the main repo. I've created a somewhat paradigm of using deno scripts for these eng scripts and I'd prefer not to add another language. If we wanted to do a scanner in this repo I imagine it would be similar to the one at https://github.com/roflmuffin/CS2-Gamedata

ELDment: fair thing, i hadn't been following that project before. since there's already a robust testing system in place there, i'll just drop this script

roflmuffin:
Thanks for all your work @ELDment, just doing a smoke test now

roflmuffin (approved):

PR #1407, opened 2026-08-30
Fix DropActivePlayerWeapon passing uninitialized vecDropMomentum

### LINKED ISSUE #1214: `CCSPlayer_ItemServices_DropActivePlayerWeapon` should be going with 2nd arg `const Vector& vecDropMomentum`
Image

Image

Image

This may lead to NaN entity velocity if `a2` were left uninitialized at that time.

I sugguest passing an empty vector to `vecDropMomentum` in `DropActiveWeapon` to keep API consistency/stable.

Umbre11as:
The vtable call for DropActivePlayerWeapon was missing the second parameter (const Vector& vecDropMomentum), so the native function read uninitialized stack for it — which could result in NaN entity velocity.

Passes an empty Vector to keep the public API stable, as suggested by @hzqst in the issue.

Tested on a live 1.0.373 server: weapon drops normally without abnormal velocity.

Closes #1214

### DIFF
--- a/managed/CounterStrikeSharp.API/Core/Model/CCSPlayer_ItemServices.cs
+++ b/managed/CounterStrikeSharp.API/Core/Model/CCSPlayer_ItemServices.cs
@@ -34,8 +34,7 @@ public partial class CCSPlayer_ItemServices
 
         Guard.IsValidEntity(activeWeapon);
 
-        var dropMomentum = new Vector();
-        VirtualFunction.CreateVoid<nint, nint, nint>(Handle, GameData.GetOffset("CCSPlayer_ItemServices_DropActivePlayerWeapon"))(Handle, activeWeapon.Handle, dropMomentum.Handle);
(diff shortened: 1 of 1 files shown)


managed/CounterStrikeSharp.API/Core/Model/CCSPlayer_ItemServices.cs
```diff
@@ -29,12 +29,13 @@ public partial class CCSPlayer_ItemServices
...
     public void DropActivePlayerWeapon(CBasePlayerWeapon activeWeapon)
     {
-        if(Handle == IntPtr.Zero)
+        if (Handle == IntPtr.Zero)
             throw new InvalidOperationException("ItemServices points to null.");
 
         Guard.IsValidEntity(activeWeapon);
 
-        VirtualFunction.CreateVoid<nint, nint>(Handle, GameData.GetOffset("CCSPlayer_ItemServices_DropActivePlayerWeapon"))(Handle, activeWeapon.Handle);
+        var dropMomentum = new Vector();
```
roflmuffin: I think we should just pin a static vector zero here rather than making a new one on each instantiation

Umbre11as: Pushed. Using Vector.Zero directly now

roflmuffin (approved):

